PR #5, opened 2018-08-14
Google-diff for enabling art-extension on Android-P celadon

yasoda-aravapalli:
Tracked-On: OAM-67515
Signed-off-by: Priyanka Bose <priyanka.bose@intel.com>


tpalli:
What is the easiest way to disable art-extension? I believe enabling it might have caused CTS build to fail. If fails with following: "error: use of undeclared identifier 'GetClass'" and GetClass is a method in 'art'.

tpalli:
> What is the easiest way to disable art-extension? I believe enabling it might have caused CTS build to fail. If fails with following: "error: use of undeclared identifier 'GetClass'" and GetClass is a method in 'art'. Or maybe there's something going wrong with art today and the patches we have on it?

found it, disabling should happen by unsetting VENDOR_ART_PATH

PR #25, opened 2018-09-06
Fix duration time incorrect by new format.

walterx-yan:
Duration time is not date format, using h:mm:ss to format it.

Tracked-On: OAM-68261


walterx-yan:
+1 tested locally,working well.

hongcheng-xie:
+1

PR #30, opened 2018-09-11
Fix the audio and video flicker problem when seeking.

walterx-yan:
1. Fix the audio and video flicker problem when seeking.
2. Filter out the dupliate call to seekTo with the same
seek time when seeking with a "touch and leave" manner.

Tracked-On: OAM-68571


hongcheng-xie:
+1  Porting from O MR1

PR #31, opened 2018-09-11
[CELADON] Modify the LensResolverActivity intent with right component

swaroopbalan:
The music can't play due to the LensResolverActivity always
use the default intent to start activity, correct it using the
ResolveInfo component.

Tracked-On: OAM-68260


walterx-yan:
No, this patch is not for OAM-68260.   OAM-68260 cannot reproduce with latest daily build. May other pathces fix it.

PR #33, opened 2018-09-12
[CELADON] Fix the time zone list is incomplete

swaroopbalan:
Add the padding between Button3(NeutralButton) and Button2(NegativeButton),
avoiding these two buttons display so closely.

Tracked-On: OAM-67832
Signed-off-by: Swaroop Balan <swaroop.balan@intel.com>


swaroopbalan:
Not needed

PR #43, opened 2018-09-19
Fix navigation bar not scaling up/down when the display size changed.

walterx-yan:
Restart the navigation bar when the density or font scale changed

Test:
1. goto Settings->Display->Advanced->Display size->Largest
2. check whether scaling up for the facet buttons of navigation bar

Tracked-On: OAM-67836


hongcheng-xie:
+1

PR #44, opened 2018-09-19
Adjust sendObject parameter's data type.

walterx-yan:
Now the MtpDevice#sendObject is using int to define object size,
if the object size is bigger than 2GB, the size will be type cast to
a negative number in MtpDevice#sendObject. And it will cause a endless
loop in MtpDataPacket#write until DUT auto reboot.

Due to the JNI api android_mtp_MtpDevice_send_object is using uint32_t
to define object size, so adjust the native api MtpDevice#sendObject
size parameter's data type to uint32_t.

Tracked-On: OAM-68165


walterx-yan:
porting from o-mr1.  @hongcheng-xie to review

hongcheng-xie:
+1

PR #47, opened 2018-09-25
[DocumentUI]Fix file cannot be accessed using mouse right click 'Open'.

arvinwax:
DocumentUI don't implement onDocumentPicked() in FilesActivity.java yet.
The solution is that sending ACTION_VIEW intent to system and system
start the right application to open the file.

Tracked-On: OAM-68477


hongcheng-xie:
+1

PR #50, opened 2018-10-10
Fix unable to input PIN password with the physical keyboard

arvinwax:
If inputting a wrong password, the Pin's input editText view will be
disabled by the verifyPassword function. Then the viewGroup will clear
the editText view's focus and find another focus view(ScrimView), it
leads to being unable to input pin password again.

Test:
1. Set a PIN from Settings->security & location->screen lock->PIN
2. Lock the device by pressing the ignition button
3. Enter a wrong PIN from a physical keyboard
4. For the next attempt, the physical keyboard cannot be used

Tracked-On: OAM-69586


hongcheng-xie:
+1

PR #52, opened 2018-10-15
Fix for Audio resumed once message is received

gaganprx:
Issue: When message readout is done, paused audio resumes
automatically.

Reason: When audio Focus changes, pausing of the audio is done
on the basis of mStreamAvailable variable. This variable was
not handled properly. It is made true at SRC_STR_START command.
When message is received, to start readout, again SRC_STR_START
is received which again makes mStreamavailable true and the
previous state of mStreamAvailable was lost. Due to which on
recepetion of AudioFocus Loss Transient, always a pause commmand
was sent and audio resumes when message readout happens.

Fix: Changed handling of mStreamAvailable from SRC_STR_START to
SRC_PLAY and SNK_PLAY.

Tracked-On: OAM-68209

Signed-off-by: Gaganpreet kaur <gaganpreetx.kaur@intel.com>


gaganprx:
@JeevakaPrabu

PR #55, opened 2018-10-17
Add new arch variant goldmont and cpu feature avx2 for cel-apl

shalinibodapati:
Tracked-On : OAM-70167
Signed-off-by: Shalini Salomi Bodapati <shalini.salomi.bodapati@intel.com>


shalinibodapati:
At this time clang  version that gets picked up at build time does not support -march=goldmont as this feature was added in later clang versions only. That's the reason we leave it empty.

jaishankintel:
Approved 

https://buildbot.sh.intel.com/absp/builders/celadon-engineering/builds/884

PR #57, opened 2018-10-19
Correct the seekbar value by min volume of the stream.

walterx-yan:
For some streams (e.g.Alarm), the min volume is not zero.
Needs to correct the value of seekbar by min volume.

Tracked-On: OAM-68651


walterx-yan:
close it due to conflict

walterx-yan:
close it due to conflict

PR #68, opened 2018-11-02
Add default values for system locales

RobinXZhang:
Now when first time log in, the system user's system locales is not
stored in settings_system database. So the value will be written by
the new user's change language action. And after switch back to system
user, the system locales value will remain the same as new user.

So the solution is add default values for system locales when system
user first time log in, and the default values is coming from device
configuration.

Tracked-On: OAM-70885
Signed-off-by: robinz1x <robinx.zhang@intel.com>


RobinXZhang:
@walterx-yan @hongcheng-xie , please review this patch, thanks.

hongcheng-xie:
+1

PR #80, opened 2018-11-12
[CEALDON]AudioPolicyManager: Avoid invalid volume setting in APM

asidda1x:
Do skip volume setting when invalid device id is sent as
argument to checkAndSetVolume().  If this setting is not skipped,
the audio volume index sent to checkAndSetVolume() may not match
with the actual audio device associated with outputDesc sent as
another argument, causing invalid volume set to actual audio device.

Tracked-On: https://jira01.devtools.intel.com/browse/OAM-71124
Signed-off-by: Aravind Siddappaji <aravindx.siddappaji@intel.com>


kkm123:
@swaroopbalan  Please merge this.

PR #88, opened 2018-11-21
[CELADON] Fix no notification seen for an alarm set

varun923:
Tracked-On:OAM-70638
Signed-off-by: varun <varunx.jeevangoudar@intel.com>


walterx-yan:
please just remove the old diff patch, no need a new diff. ::)

PR #89, opened 2018-11-21
IVI: Set sys.usb.config in suspend/resume

saranyagopal1:
This adds a patch to set sys.usb.config
during suspend and resume.

Jira: OAM-71949

Signed-off-by: Saranya Gopal <saranya.gopal@intel.com>


saranyagopal1:
Should be done in powermanagement service.

PR #90, opened 2018-11-22
Remove the diff of DeskClock as P already fix the original bug.

walterx-yan:
The old diff patch is for O-MR1, and not needed for P.
And it will cause a regression, so remove it.

Tracked-On: OAM-70638


hongcheng-xie:
+1

PR #94, opened 2018-11-23
Intel BT Controller can not handle 2 Commands from the host

goswamiharshita:
at the same time resulting in unstatable Bluetooth.

Fix the issue by reducing the command credit in all cases equal
to 1.

Pairing/Connection/File Transfer/A2DP found to be working.

Tracked-On: OAM-71561
Signed-off-by: Atul Vaish <atul.vaish@intel.com>
Signed-off-by: Jeevaka Prabu Badrappan <jeevaka.badrappan@intel.com>


goswamiharshita:
@JeevakaPrabu

PR #96, opened 2018-11-27
Add ddb algorithm optimization patch to fix wm issue

yugang:
Tracked-On: OAM-72227
Signed-off-by: Fan Yugang <yugang.fan@intel.com>

hold on and need verification, then merge, thanks


renchenglei:
@yugang @swaroopbalan with above patch, the WM related issue can't be reproduced any more!

PR #105, opened 2018-12-03
Check if mClingWindow is already added before adding.

walterx-yan:
If not to check if it's added, the previous one will leak
and cannot be removed any more.

Tracked-On: OAM-7240


walterx-yan:
@hongcheng-xie to review. Thanks.

PR #114, opened 2018-12-05
Resuming display after S3

smadhusu:
On platforms like KBL, the display was never resuming
on single power button press. This work-around will fix
issues where single power button press is not detected
during wakeup.

Tracked-on: OAM-72346


smadhusu:
Hi @abhilashkesavan please review the patch for resuming display from S3.

PR #119, opened 2018-12-07
Fix for BT reset issue

aiswaryacyriac:
Intel BT Controller can not handle 2 Commands from the host
at the same time resulting in unstatable Bluetooth.

Fix the issue by reducing the command credit in all cases equal
to 1.

Test: Pairing/Connection/File Transfer/A2DP found to be working.

Tracked-On: OAM-72658
Signed-off-by: Aiswarya Cyriac <aiswarya.cyriac@intel.com>


aiswaryacyriac:
@JeevakaPrabu @goswamiharshita

PR #121, opened 2018-12-10
Fix "Guest" user option seen when logged in as "New User"

arvinwax:
if the foreground user is the device owner, add the guest
user record.

Tracked-On: OAM-70634


arvinwax:
Verify +1

arvinwax:
Hi @hongcheng-xie @walterx-yan 

PR #126, opened 2018-12-12
[CELADON] Bug-fix: Screen will not display

smadhusu:
“Power/Restart/Sleep” after S3.

Screen will not display “Power/Restart/Sleep”
options after resume from S3. However this issue
was specific to APL.

Tracked-On: OAM-72731


smadhusu:
@abhilashkesavan Please review the changes for bug-fix..

PR #130, opened 2018-12-13
Fix testScreenBrightness failed in CtsStatsdHostTestCases Module

arvinwax:
Case fail due to update a wrong backlight brightness to the Setting database.
Due to the display brightness for the vehicle, need a percent brightness.
but the System screen backlight brightness is a value between 0 and 255, need
to convert the screen backlight brightness to the percent brightness. If the
screen backlight brightness had changed from the vehicle, need to convert the
percent brightness to screen backlight brightness value, So unable to save a
accurate brightness value to the Setting database.

Test:
1. adb shell settings put system screen_brightness 100
2. adb shell settings get system screen_brightness

Tracked-On: OAM-73451


arvinwax:
Verify +1

arvinwax:
Hi @walterx-yan , @hongcheng-xie 
thank you.

PR #133, opened 2018-12-17
[CELADON] Screen will not display “Power/Restart/Sleep”

smadhusu:
Screen will not display “Power/Restart/Sleep”
options after resume from S3. However this issue
was specific to APL.

Tracked-On: OAM-72731


smadhusu:
Hi @abhilashkesavan please review the patch for display bug-fix after resume..

PR #135, opened 2018-12-18
[WA]Decrease the duration of screen-off animation to get CTS pass.

walterx-yan:
testWakelockState requires screen off before it's running.
But the current source code of case testPartialWakelock does not
guarantee that. This workaround will get screen off faster to
get the case pass.

Tracked-On: OAM-72680


walterx-yan:
@hongcheng-xie to review. Thanks.

PR #143, opened 2018-12-27
Issue: Fix for Audio not routing to DUT after BT OFF/ON

gaganprx:
Reference phone Audio over Bluetooth is not routing
to IVI speakers after bluetoth turn OFF and ON from DUT car
settings.

Analysis: the root cause of the issue is as:

1. After BT OFF/ON from the Car Settings, when BT turns on,
it is initiating the scanning and also at the same time,
the connection to the already bonded devices starts.
2. As the controller is performing discovery of devices, so
when it receives "Create Connection" Command, it rejects the
command with the status "Command Disallowed- 0x0c".
3. Due to this, A2DPSink and HEADSET_CLIENT profiles are not
able to connect.

Fix: Removed StartScanning when BT turns on in Car Settings.
Scanning will not be initiated automatically on BT ON event
from Car Sttings App.

Test: BT Pairing, Audio routing after BT OFF/ON working fine

Tracked-On: OAM-72646


gaganprx:
@JeevakaPrabu Please review

raveendra-chennakesavulu:
@swaroopbalan @sgnanase, please merge this patch

PR #209, opened 2019-02-12
Fix bug of Notification Listener Test

walterx-yan:
To save the settings even it's empty.

Tracked-On: OAM-76029


walterx-yan:
@swaroopbalan to review. Thanks.

PR #224, opened 2019-02-20
Android Enabling Changes in OPENIVNO

avaish1:
Tracked-On: OAM-76180
Signed-off-by: Atul Vaish <atul.vaish@intel.com>


swaroopbalan:
it has to be merged in openvino, owned by other intel team--> update from Atul

PR #227, opened 2019-02-21
Avoids going to blank screen when Max users are created & Fix the behavior of keyguard bouncer in a corner case

varun923:
Tracked-On: OAM-76010
Signed-off-by: varun <varunx.jeevangoudar@intel.com>


varun923:
Thanks Walter

PR #265, opened 2019-03-25
The path-permission element offers prefix or regex style matching of

arvinwax:
paths, but most providers internally use UriMatcher to decide what
to do with an incoming Uri.

This causes trouble because UriMatcher uses Uri.getPathSegments(),
which quietly ignores "empty" paths.  Consider this example:

    <path-permission android:pathPrefix="/private" ... />

    uriMatcher.addURI("com.example", "/private", CODE_PRIVATE);

    content://com.example//private

The Uri above will pass the security check, since it's not
technically a prefix match.  But the UriMatcher will then match it
as CODE_PRIVATE, since it ignores the "//" zero-length path.

Since we can't safely change the behavior of either path-permission
or UriMatcher, we're left with recovering these shady paths by
trimming away zero-length paths.
cherry-pick from:
https://android.googlesource.com/platform/frameworks/base/+/798aba1%5E%21/

Bug: 112555574
Test: atest android.appsecurity.cts.AppSecurityTests
Test: atest FrameworksCoreTests:android.content.ContentProviderTest

Tracked-On: OAM-79047

### DIFF
--- a/android_p/google_diff/cel_apl/frameworks/base/0028-Recover-shady-content-paths.patch
+++ b/android_p/google_diff/cel_apl/frameworks/base/0028-Recover-shady-content-paths.patch
@@ -1,0 +1,258 @@
+From 35f31cbac405b20ec1cd55714609c88f98a99c14 Mon Sep 17 00:00:00 2001
+From: Jeff Sharkey <jsharkey@android.com>
+Date: Mon, 24 Sep 2018 13:23:57 -0600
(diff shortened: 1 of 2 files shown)


arvinwax:
Hi @hongcheng-xie @walterx-yan 
Verify +1

Thank you.

hongcheng-xie:
Arvin, it's better to add google original patch link info in the commit message.

PR #302, opened 2019-04-30
Kernelflinger: Disable NVMe RPMB supported in WHL

zhouyanhongx:
In WHL RVP, the NVMe RPMB is not supported and cause flash failed.
But we also need to support eMMC RPMB using one binary.
So we need to disable the NVMe RPMB suuport in clk lunch target.

Tracked-On: OAM-80029
Signed-off-by: Yanhongx.Zhou <yanhongx.zhou@intel.com>


tanminger:
> directly comment code seems not good way

If delete the code, then will cause the code line number changed, and easy to cause apply patchs conflict.

PR #336, opened 2019-05-31
Fix is to check cpu_freq for 0 value.

zhouyanhongx:
RDMSR returns 0 in VM thus causing divide by error below.  Fix is to
check cpu_freq for 0 value and set a flag "time_stamp = FALSE" and
exit the boot time calculation.

Tracked-On: OAM-80206
Signed-off-by: Yanhongx.Zhou <yanhongx.zhou@intel.com>


tanminger:
Replace with https://github.com/projectceladon/vendor-intel-utils/pull/350

PR #337, opened 2019-05-31
test suspend/resume fix

saranyagopal1:


balajim001:
+1 , Looks good from USB side

PR #353, opened 2019-06-14
[DO NOT MERGE] This is to align with mesa 19.0.6

renchenglei:
Please don't merge this changes, we keep it here to align with
new mesa 19.0.6 to pass opensource gfx master building pass


tpalli:
FYI the '_pri' changes should also include Iris driver. Makefile is src/gallium/drivers/iris/Android.mk and there libdrm_intel needs to be changed to libdrm_intel_pri when building Iris on Celadon. Maybe you could update the original patch to cover this?

renchenglei:
> FYI the '_pri' changes should also include Iris driver. Makefile is src/gallium/drivers/iris/Android.mk and there libdrm_intel needs to be changed to libdrm_intel_pri when building Iris on Celadon. Maybe you could update the original patch to cover this?

Thanks for the info. Sure, I will update the original patch to cover this, :)

PR #362, opened 2019-06-24
usb: dbc: adb root

revatiranjan:
adb root for KBL


revatiranjan:
not required , push other changes for same issue.

PR #374, opened 2019-07-04
Sanitize InputMessage before sending

arvinwax:
The struct InputMessage has many fields, and is force-aligned to 8-byte
boundaries. There are also some padding fields that carry no
information.

This struct is typically allocated in the stack and populated with
various values before being sent across as a stream of bytes through the
socket.

Therefore, the "unused" data portions of the struct could contain
portions of the stack, since there aren't ever writes to those memory
locations.

To avoid this information leak, forcefully sanitize the struct. Create a
new struct that is explicitly set to zero. Next, only fill the
meaningful fields manually.

Bug: 115739809
Test: cts-tradefed run cts -m CtsSecurityBulletinHostTestCases -t android.security.cts.Poc18_12; adb shell monkey 100000
Change-Id: I7e44dacf1e8fa3156c8e4d2f7784ef0c53dab507
Merged-In: I7e44dacf1e8fa3156c8e4d2f7784ef0c53dab507
(cherry picked from commit e730f5aaa1c726ee9998a080e2d7f6284f4afec8)

Tracked-On: OAM-83600
Signed-off-by: johnzhax <johnx.zhang@intel.com>


rayx-lei:
@hongcheng-xie please help review it.

PR #376, opened 2019-07-08
Add support call disconnect cause DisconnectCause.REJECTED in HFPClient

arvinwax:
When a phone call disconnected by the user rejected, make sure that it
goes into the rejective call log rather than the missing call log.

Change-Id: Id252ef4432b3bc80e1c9b25185732d310a413846
Tracked-On: OAM-82986


arvinwax:
verify +1
Hi @hongcheng-xie 
Please help to reveiw.

PR #402, opened 2019-08-06
Disable RBC on cel_apl and celadon

yhe39:
The RBC will cause 4K display stuck

Test: 4K display works well, no regression.
Tracked-On: OAM-84580
Signed-off-by: HeYue <yue.he@intel.com>

### DIFF
--- a/android_p/google_diff/cel_apl/vendor/intel/external/project-celadon/hwcomposer/0001-Disable-RBC.patch
+++ b/android_p/google_diff/cel_apl/vendor/intel/external/project-celadon/hwcomposer/0001-Disable-RBC.patch
@@ -1,0 +1,34 @@
+From 2fcc2da1fe97ad197ec99313009861e649c37c94 Mon Sep 17 00:00:00 2001
+From: HeYue <yue.he@intel.com>
+Date: Fri, 24 May 2019 09:42:36 +0800
+Subject: [PATCH] Disable RBC
+
+On clk, if connect two displays, 4K as primary display. The RBC
+will cause 4K display stuck.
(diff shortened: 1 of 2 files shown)


yhe39:
@YuanjunHuang, please help to review. Thanks!

tpalli:
Note, this has a direct impact to performance benchmarks. Would it be possible to disable rbc dynamically when 4K display is found (and not always)?

yhe39:
hi @tpalli ,this patch just disable rbc in hwc, the rbc feature still exist in kernel, we checked that if disable rbc in hwc, the performance do not have much impact. thanks!

tpalli:
@yhe39  rbc still exists in kernel but unfortunately gets unused if the userspace does not utilize it. I'm surprised to hear it does not have much impact, on desktop Linux the impact is quite big, perhaps the RBC pipeline was not really working as it should.

YuanjunHuang:
@tpalli yes actually in Android we did not see convicing performance gain with RBC but we do see many flicker issues with RBC. we enabled RBC in hwcomposer only for render targets that generated in hwc for composing layers that are not directly scanned out due to hw plane limitation. Now our implementation has changed to compose most of these layers by android framework (surfaceflinger) which has not yet implemented RBC. So the impact is not big. Given the tradeoff of current status, we decide to disable RBC for now.     maybe we can investigate 'the RBC pipeline was not really working as it should.' as you said in the future. thanks.

yhe39:
hi @YuanjunHuang, could u help to review this patch?
Thanks!

YuanjunHuang:
@yhe39  please check with Jenny if it also needs to submit into q-migration branch.

PR #457, opened 2019-09-16
[CELADON] HDMI plug/unplug events are not reported using getevent iss…

swaroopbalan:
…ue is seen in Commercial KBL NUC

Tracked-On: OAM-86570
Signed-off-by: swaroopb <swaroop.balan@intel.com>


yjsun1:
This patch applied successfully on targets: caas, celadon_ivi, celadon_tablet.

celadon EB #4276

Project kernel/lts2018
	Applying		0001-CELADON-HDMI-plug-unplug-events-are-not-reported-usi.patch

PR #498, opened 2019-09-27
Remove unused boot bsp_diff patch

zjin7:
Tracked-On: OAM-86869
Signed-off-by: Zhi Jin <zhi.jin@intel.com>


zhimincx:
EB4633 test pass

PR #521, opened 2019-10-12
Set config_quickSettingsSupported to false for IVI

shellyxxxie:
No quickSettings on IVI system, so disable it.

Tracked-On: OAM-88196
Signed-off-by: Xie, Shellyx <shellyx.xie@intel.com>


shellyxxxie:
@walterx-yan , @hongcheng-xie ,
Please review this commit.

PR #569, opened 2019-11-13
bsp_diff: Add change to fix USB gadget crash fix

saranyagopal1:
Add kernel patch to fix crash while switching
between multiple USB gadget functions.

Signed-off-by: Saranya Gopal <saranya.gopal@intel.com>


saranyagopal1:
Will come from PKT.

PR #587, opened 2020-01-02
rebase mesa bsp diff for 19.1.6

renchenglei:


renchenglei:
When we upgrade mesa to 19.1.6, we may encounter conflict with bsp diff, this changes helps to fix the conflicts issue.
The 0002 patch has been included in 19.1.6. So we don't need it any more.
commit f47405f3a1b3d157fb128eeab34487183feaf7d6
Author: Alistair Strachan <astrachan@google.com>
Date:   Wed Sep 19 15:05:33 2018 -0700

    UPSTREAM: mesa: android: Remove unnecessary dependency tracking rules

    The current AOSP master build system breaks building mesa due to the
    following error:

    external/mesa3d/src/compiler/Android.glsl.gen.mk:94: error:
      writing to readonly directory: "external/mesa3d/src/compiler/glsl/ir.h"

    This error is bogus -- nothing "writes" to ir.h -- but the rule is
    unnecessary because the generated header that is a dependency of the
    non-generated header should be added to LOCAL_GENERATED_SOURCES and this
    will track if the dependency needs to be regenerated.

renchenglei:
@YuanjunHuang, @feijiang1, @hongcheng-xie, please help take a review of this changes.
@hongcheng-xie, I have no yifei and xie, chao's ID, could you help check this? Thanks a lot!

PR #602, opened 2020-03-11
Remove selinux related mount option since selinux is disabled

Zhiwei-Lii:
Tracked-On: OAM-89794
Signed-off-by: Zhiwei Li <zhiwei.li@intel.com>


hongcheng-xie:
zhiwei, selinux will be enabled later in cic target, pls move the patch under cic_dev/.

PR #640, opened 2020-04-21
enable passthrough virtio-input device

dyang23:
Enable virtio-input device for multi-touch device.
For kernel lts2018, we use a passthrough patch.
For kernel lts2019-yocto, we back ported a patch from Google.

Tracked-On: OAM-90659
Signed-off-by: Yang, Dong <dong.yang@intel.com>


dyang23:
LGTM and both patch test passed.

PR #663, opened 2020-04-30
Let some processes of cic run as system user

ceiba1985:
For security reason, our customer want all the processes of
a container to run as non-root user, so let some processes
that run as root user previously run as system user.

Tracked-On: OAM-90885

### DIFF
--- a/android/cic/bionic/0001-Remove-uid-gid-check-for-device-property-files.patch
+++ b/android/cic/bionic/0001-Remove-uid-gid-check-for-device-property-files.patch
@@ -1,0 +1,32 @@
+From 8fa1f7da877180258f18e47d131f76af37d4464c Mon Sep 17 00:00:00 2001
+From: "ji, zhenlong z" <zhenlong.z.ji@intel.com>
+Date: Thu, 30 Apr 2020 14:09:30 +0800
+Subject: [PATCH] Remove uid/gid check for device property files
+
+Basing on the need of rootless process, device property files
+maybe created by a non-root process, this uid/gid check is
+improper for this scenario.
+
+Change-Id: Ie24b5fd0ff05caa81a40982a7fe409e61dac8946
+Signed-off-by: ji, zhenlong z <zhenlong.z.ji@intel.com>
+---
+ libc/system_properties/prop_area.cpp | 3 +--
+ 1 file changed, 1 insertion(+), 2 deletions(-)
+
+diff --git a/libc/system_properties/prop_area.cpp b/libc/system_properties/prop_area.cpp
+index 42bee9f3e..c0ab5af39 100644
+--- a/libc/system_properties/prop_area.cpp
++++ b/libc/system_properties/prop_area.cpp
+@@ -110,8 +110,7 @@ prop_area* prop_area::map_fd_ro(const int fd) {
+     return nullptr;
+   }
+ 
+-  if ((fd_stat.st_uid != 0) || (fd_stat.st_gid != 0) ||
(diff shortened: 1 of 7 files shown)


android/common/system/core/0005-Let-some-core-services-run-as-system-user.patch
```diff
@@ -0,0 +1,290 @@
...
+         if (prctl(PR_SET_SECUREBITS, securebits) != 0) {
+             PLOG(FATAL) << "prctl(PR_SET_SECUREBITS) failed for " << name_;
+         }
+@@ -916,6 +919,11 @@ Result<Success> Service::Start() {
+             ZapStdio();
+         }
+ 
++        if (strcmp("adbd", name_.c_str()) == 0) {
++            if (getuid() != 0)
++                setuid(0);
```
joez, line 179: which means adbd will still run as root?

ceiba1985: adbd will invoke setuid itself. consider adb root/unroot.

android/common/system/core/0005-Let-some-core-services-run-as-system-user.patch
```diff
@@ -0,0 +1,290 @@
...
+
+diff --git a/init/builtins.cpp b/init/builtins.cpp
+index 8bd92ccdd..147069f29 100644
+--- a/init/builtins.cpp
++++ b/init/builtins.cpp
+@@ -128,9 +128,21 @@ static Result<Success> do_class_restart(const BuiltinArguments& args) {
+ }
+ 
+ static Result<Success> do_domainname(const BuiltinArguments& args) {
++    int ori_uid = getuid();
```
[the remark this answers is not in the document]
ceiba1985: Can do this, it just modify an API may introduce more modifications.

android/common/system/core/0005-Let-some-core-services-run-as-system-user.patch
```diff
@@ -0,0 +1,319 @@
...
+     android::base::unique_fd fd(TEMP_FAILURE_RETRY(
+         OpenFile(path, O_WRONLY | O_CREAT | O_NOFOLLOW | O_TRUNC | O_CLOEXEC, 0600)));
+     if (fd == -1) {
++        if (ori_uid > 0 && as_root)
++            setuid(ori_uid);
+         return ErrnoError() << "open() failed";
+     }
+     if (!android::base::WriteStringToFd(content, fd)) {
++        if (ori_uid > 0 && as_root)
++            setuid(ori_uid);
```
joez, line 194:
```cpp
SETUID_END;
```

android/common/system/core/0005-Let-some-core-services-run-as-system-user.patch
```diff
@@ -0,0 +1,319 @@
...
+         return ErrnoError() << "Unable to write file contents";
+     }
++    if (ori_uid > 0 && as_root)
++        setuid(ori_uid);
```
joez, line 198:
```cpp
SETUID_END;
#undef SETUID_START
#undef SETUID_END
```

joez:
a better way to achieve this is create a wrapper function, e.g.

```cpp
static bool SetHighestAvailableOptionValueAsRoot(std::string path, int min, int max) {
  int ori_uid = 0;
  if (as_root && (ori_uid = getuid()) != 0)
      setuid(0);
  bool ret = SetHighestAvailableOptionValue(path, min, max);
  if (as_root && ori_uid != 0)
      setuid(ori_uid);
  return ret;
}
```

joez:
LGTM

PR #725, opened 2020-06-04
Add a item in get_param IOCTL for query GVT active

Shao-Feng:
When GVT is active, return true.

Tracked-On: OAM-90744


Shao-Feng:
Hi @JeevakaPrabu 
Please review it. Thanks

PR #729, opened 2020-06-05
Add a item in get_param IOCTL for chromium kernel GVT active

Shao-Feng:
When GVT in chromium kernel is active, return true.
    that is:
    GVT-g --- true
    GVT-d --- false
    native --- false

    Tracked-On: OAM-90744


Shao-Feng:
> * chromium patch number has to be updated to 12
> * As yocto patch is already merged, please remove it from this patch

Updated on
https://github.com/projectceladon/vendor-intel-utils/pull/730

Please review #730. I will close this one

PR #730, opened 2020-06-08
Add a item in get_param IOCTL for chromium kernel GVT active

Shao-Feng:
When GVT in chromium kernel is active, return true.
that is:
GVT-g --- true
GVT-d --- false
native --- false

Tracked-On: OAM-90744


Shao-Feng:
Thanks, I will update.

Shao-Feng:
Hi @JeevakaPrabu 
It is updated. Please help to review. thanks a lot to your comments

PR #741, opened 2020-06-12
limit the num of patches we checked the applied patches

yyao3:
set the num as 10000 for now

Tracked-On: OAM-91414
Signed-off-by: Yong Yao <yong.yao@intel.com>


sgnanase:
Good One (y) !!

yyao3:
> Good One (y) !!

Thank you. what's the next step to get this merged?

PR #760, opened 2020-06-28
Fix for WiFi status is enabled default.

raveendra-chennakesavulu:
On very first boot, though wlan is down wifi status in Settings app showing
as enabled.

Tracked-On: OAM-91559
Signed-off-by: Raveendra Babu Chennakesavulu
                        <raveendra.babu.chennakesavulu@intel.com>


raveendra-chennakesavulu:
pls review @JeevakaPrabu

PR #862, opened 2020-09-28
Fix signapk.jar not found in RELEASE_BUILD=true

tanujtekriwal:
When built with RELEASE_BUILD=true option builder tries to
call the signapk.jar file to generate the caas-sign***.zip file.
This signapk path is not set properly in Android 11.
The path is set relative to the script directory but it should
be relative to the output directory.
This patch fixes this issue and sets the path properly.

### DIFF
--- a/aosp_diff/caas/build/make/0003-Fix-signapk.jar-path-for-RELEASE_BUILD-true-case.patch
+++ b/aosp_diff/caas/build/make/0003-Fix-signapk.jar-path-for-RELEASE_BUILD-true-case.patch
@@ -1,0 +1,34 @@
+From 7289608cee0cd5edaf7a9a9a0eeb9e99f0b81748 Mon Sep 17 00:00:00 2001
+From: Tanuj Tekriwal <tanuj.tekriwal@intel.com>
+Date: Mon, 28 Sep 2020 20:39:45 +0530
+Subject: [PATCH] Fix signapk.jar path for RELEASE_BUILD=true case
+
+This patch fixes the issue with signapk.jar path.
+This path was set to the script directory but needed to be
+set to the output directory.
+Without this if code built with RELEASE_BUILD=true will fail.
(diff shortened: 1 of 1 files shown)


shyjumon-n (approved):
Looks good.

aosp_diff/caas/build/make/0003-Fix-signapk.jar-path-for-RELEASE_BUILD-true-case.patch
```diff
@@ -0,0 +1,34 @@
...
+diff --git a/tools/releasetools/common.py b/tools/releasetools/common.py
+index ae7da2fb6..8726609f4 100644
+--- a/tools/releasetools/common.py
++++ b/tools/releasetools/common.py
+@@ -62,7 +62,7 @@ class Options(object):
+           'Warning: releasetools script should be invoked as hermetic Python '
+           'executable -- build and run `{}` directly.'.format(script_name[:-3]),
+           file=sys.stderr)
+-    self.search_path = os.path.realpath(os.path.join(os.path.dirname(exec_path), '..'))
++    self.search_path = os.path.realpath(os.path.join(os.path.dirname(exec_path), '../../../../out/host/linux-x86'))
```
JeevakaPrabu, line 28: Is there any pre-defined macros we can use instead of ../../../../out/host/linux-x86 ?

tanujtekriwal: No there isn't. In Android Q this search_path was handled differently by starting from"out" and then adding "host/linux-x86" or "darwin-x86" based on the build system. Now they have changed it like this and starting from workdir. We will be raising bugizer ticket with google to discuss the same. 
As of now this will unblock SI / testing team.

JeevakaPrabu (approved):

PR #890, opened 2020-10-13
WA patch to enable adb in android user build.

tanujtekriwal:
This patch enables the adb in user build of android
by bypassing the authentication mechanism.

Tracked-On: OAM-93155


shyjumon-n (approved):
Disabling auth to unblock activities. Proper solution need to be merge later.

Approved.

PR #923, opened 2020-11-09
Filterout isa perf libraries from vndk list

amritaintel:
Currently we have added perf libs under vndksp,
filter them out as they are not new libraries,
rather avx2 implementations of existing libraries.
Should have no impact on GSI.

Tracked-On: OAM-94541


vinaykompella (approved):
This should not have any impact on GSI. Looks good to me.

qshi5 (approved):

PR #936, opened 2020-11-25
Skip CFI checking for 2 comparator

Shao-Feng:
The function pointer "void * priv" are defined in <linux/list_sort.h>
we can not change it to any other type, althouth this pointer is not
used at all.

Tracked-On: OAM-94511 and OAM-94603


Shao-Feng:
The patch did not work, even adding __nocfi in HEAD files

PR #945, opened 2020-12-11
Enable AIDL Interface for Lights HAL

SaliniVenate:
Signed-off-by: svenate <salini.venate@intel.com>


shyjumon-n (approved):
Looks good to me.

PR #968, opened 2021-01-12
Fix MultiDisplaySecurityTests failure issue

renchenglei:


renchenglei:
@feijiang1 & @YuanjunHuang & @JeevakaPrabu 
Could you help take a review of this changes? Thanks a lot!

feijiang1 (commented):
we need check why GL_CHECKPOINT(LOW) failed. Should glGetError() return false, then need check which caused GL failure before calling swapbuffer.

renchenglei:
Thanks Fei! Yes, it is! I will hold on this changes, and do more investigation.

PR #992, opened 2021-02-08
Diff for setting the fixed performance mode

shwethabolumbu:
This will be removed when Power hal Aidl is
moved to vendor/intel/hardware/interfaces

Currently in AIDL, only fixed performance mode
is enabled. Other mode/Hints are handled in
Power hal directly.

Tracked-On: OAM-95866
Signed-off-by: Shwetha B <shwetha.b@intel.com>


kaushlen (approved):

JeevakaPrabu (approved):

ceiba1985 (approved):

JeevakaPrabu (approved):
Trigger build with make options to validate sepolicy complaince too

PR #996, opened 2021-02-10
camera service remove

shivasku82:


JeevakaPrabu (changes requested):
include package android.hardware.camera.provider@2.4-service_64 instead of removing the 32bit binary in product.mk

PR #1027, opened 2021-03-01
 AOSP Patch to fix cts issue

SaliniVenate:


shyjumon-n (approved):
Looks good to me

PR #1039, opened 2021-03-08
return from functions when directory does not exist. Lets me apply patches successfully without syncing the whole tree. Otherwise adds files in wrong places...

seware:
Signed-off-by: seware <stephen.e.ware@intel.com>


seware:
Cancel this one in favor of next one....

PR #1094, opened 2021-05-06
sepolicy: allow system_server to read /proc/locks

SaliniVenate:
Access to /proc/locks is necessary to activity manager to determine
wheter a process holds a lock or not prior freezing it.

Tracked-On: OAM-96736


shyjumon-n (approved):
Looks good to me.

PR #1125, opened 2021-06-29
Removing patches which are not neeeded after

swaroopbalan:
taking the latest AOSP tag android-11.0.0_r38

Tracked-On: OAM-97430
Signed-off-by: swaroopb <swaroop.balan@intel.com>


shyjumon-n (commented):
Should we change the below commit msg:

taking the latest AOSP tag android-11.0.0_r38
to something like:
Rebasing to latest AOSP tag android-11.0.0_r38

shyjumon-n (approved):
LGTM

PR #1170, opened 2021-08-25
Enable adb in user build by default for caas_cfc

ceiba1985:
Looking Glass of PGP has dependence on adb. In user build
the adb is disabled by default, so the UI cannot be displayed
by LG. And to turn on adb, we need to access the UI, we are
stuck in a deadlock.

Tracked-On: OAM-98789


JeevakaPrabu (approved):

shyjumon-n (approved):
Looks Good to me

swaroopbalan (approved):

PR #1187, opened 2021-09-10
Hide status bar and grant runtime permission by default

ceiba1985:
Based on the customer requirement, we need to remove the
status bar and disable the permission grant pop-up window.

Tracked-On: OAM-96162


JeevakaPrabu (approved):

shyjumon-n (approved):
Looks good to me

PR #1236, opened 2021-10-28
WA for app orientation issue

gkdeepa:


aosp_diff/caas_cfc/frameworks/base/0004-WA-for-rotation-framework.patch
```diff
@@ -0,0 +1,114 @@
...
+ struct fields_t {
+     jfieldID    context;
+     jfieldID    facing;
+@@ -544,9 +544,14 @@ static void android_hardware_Camera_getCameraInfo(JNIEnv *env, jobject thiz,
+         jniThrowRuntimeException(env, "Fail to get camera info");
+         return;
+     }
++    if(checkflag == 1){
++       if(cameraInfo.facing == 1) 
++        env->SetIntField(info_obj, fields.orientation, 270);
```
renchenglei, line 30: Take Classin as one example, when the front camera is called, then we need "SetIntFied", but when it is back, we don't need call "SetIntFied"?

PR #1276, opened 2021-11-19
Downgrade the SPL to match that of android-11.0.0_r48

SaliniVenate:
Tracked-On:


shyjumon-n (approved):
LGTM

PR #1285, opened 2021-11-25
update policy for 3rd party App function

ruanhongfu:
policy update here is to allow sepolicy update under device/intel/sepolicy/aafd,
or build process would be broken

Tracked-On: OAM-100126
Signed-off-by: Ruan, Hongfu <hongfu.ruan@intel.com>


shyjumon-n (approved):
LGTM

PR #1312, opened 2021-12-13
Set Audio buffer size in guest

padmashree9110:
change period size in guest according in host
Signed-off-by: pmandri <padmashree.mandri@intel.com>


gkdeepa (approved):
Changes added for PGP to address glitch issue randomly during playback.

sgnanase (approved):

PR #1331, opened 2022-01-11
adjust system policy to allow houdini changes

ruanhongfu:
these changes are necessary to pass compile process with houdini changes

Tracked-On: OAM-100437
Signed-off-by: Ruan, Hongfu <hongfu.ruan@intel.com>


ceiba1985 (commented):
Is this patch for caas_cfc only? if yes, can you move this patch to vendor/intel/utils/aosp_diff/caas_cfc?

PR #1338, opened 2022-01-18
Adding Google patch to fix STS issue

tanujtekriwal:
This patch is available in Android 12 and
not available in Android 11. But it is needed
for fixing STS issue:
TC:  android.security.sts.CVE_2021_0310#testPocCVE_2021_0310

Tracked-On: OAM-100715


shyjumon-n (approved):
LGTM

PR #1341, opened 2022-01-19
Enable virtual ab compression ota

ceiba1985:
Need this patch to enable virtual ab compression feature,
or the system sometimes cannot boot up after the OTA completes.

Simply speaking, snapuserd is needed to be started before
the init second stage, since system partition need to be
setup after the snapuserd is started, or the system partition
cannot be accessed. However, init second stage needs to
access the system partition, So add this patch to make sure
snapuserd is started before init second stage.

Will report a bug to Google, and this patch also will be removed
once this issue is resovled.

Tracked-On: OAM-100724

### DIFF
--- a/aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
+++ b/aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
@@ -1,0 +1,120 @@
+From d549b58fa8253765a78414c09ba0fb79168e7ab7 Mon Sep 17 00:00:00 2001
+From: "ji, zhenlong z" <zhenlong.z.ji@intel.com>
+Date: Tue, 18 Jan 2022 16:16:26 +0800
+Subject: [PATCH] Let snapuserd started before init second stage
+
+1. snapuserd need to be put to recovery/root/first_stage_ramdisk/system/bin
+when BOARD_USES_RECOVERY_AS_BOOT := true is set, or it will appear the
+following errors:
+[   11.602223] init: Relaunched snapuserd with pid: 152
+[   11.602700] init: Creating logical partitions with snapshots as needed
+[   11.603555] init: Cannot launch snapuserd; execv failed: No such file or directory
+
+2. snapuserd need to be started before the init second stage, or it will
+appear the following errors. It because if init second stage is started before snapuserd,
+/dev/__properties__ will be created by init. Then in the function of
+__system_properties_init during the libc init stage of snapuserd process, it will
+go to ContextsSplit.Initialize finally, and in this function, it will try to read
+/system/etc/selinux/plat_property_contexts. However, system partition should be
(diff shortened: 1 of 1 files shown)


aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
```diff
@@ -0,0 +1,120 @@
...
+diff --git a/fs_mgr/libsnapshot/snapuserd_daemon.cpp b/fs_mgr/libsnapshot/snapuserd_daemon.cpp
+index 7fa01b78f..ddfd1ed1c 100644
+--- a/fs_mgr/libsnapshot/snapuserd_daemon.cpp
++++ b/fs_mgr/libsnapshot/snapuserd_daemon.cpp
+@@ -122,6 +122,8 @@ int main(int argc, char** argv) {
+ 
+     android::snapshot::Daemon& daemon = android::snapshot::Daemon::Instance();
+ 
++    LOG(INFO) << "Snapuserd daemon start.";
++    close(open("/dev/.snapuserd", O_WRONLY | O_CREAT | O_CLOEXEC, 0000));
```
GangSecurity, line 76: if open returns a invalid fd, then close with this fd, is there any issue?

aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
```diff
@@ -0,0 +1,120 @@
...
++#include <thread>
+ 
+ #include <android-base/file.h>
+ #include <android-base/logging.h>
+@@ -247,6 +248,14 @@ void SnapuserdSelinuxHelper::RelaunchFirstStageSnapuserd() {
+ 
+         setenv(kSnapuserdFirstStagePidVar, std::to_string(pid).c_str(), 1);
+ 
++        while(access("/dev/.snapuserd", F_OK) != 0) {
++            std::this_thread::sleep_for(100ms);
```
GangSecurity, line 97: Is there a issue if this is a infinite loop?

aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
ceiba1985: EBADF will be returned. Actually we reference the code here:
https://cactus-source.iind.intel.com/source/xref/CELADON-MASTER-R2/system/core/init/init.cpp#785

aosp_diff/preliminary/system/core/02_0002-Enable-virtual-ab-compression-feature.patch
ceiba1985: We must ensure snapuserd is started before moving to the next steps, or the system will hang. So wait for this file in the infinite loop.

GangSecurity (approved):

PR #1345, opened 2022-01-21
Patch to disable unneeded services/packages

tanujtekriwal:


JeevakaPrabu (changes requested):
Are this properties set in mk file?

JeevakaPrabu (approved):

PR #1357, opened 2022-02-03
Add SOF firmware and topology files

padmashree9110:
This patch adds sof firmware and topology files for ADL
Signed-off-by: pmandri <padmashree.mandri@intel.com>


padmashree9110:
Creating new PR with correction

PR #1397, opened 2022-03-23
[VSOCK] Support for Suspend & Resume.

vilasrk:
[VIRTIO] Freeze and Restore Implementation.

Whenever VM goes for suspend if there are any
active VSOCK connections used to communicate with host,
respective vsock connections were unable to resume as
freeze and restore functionality was not implemented in
virtio driver.
This change will fix VSOCK reconnect in case
of Suspend and Resume

 In freeze case:
    Change will reset all connected sockets when the device
    disappears and flush all device writes and interrupts
    and also delete virtqueues and flush outstanding
    callbacks if any.

In Restore case:
    Change will initialize all the necessary parameter to
    make the virtio device ready

Tracked-On: OAM-100107
Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>
                        Vilas R K <vilas.r.k@intel.com>
                        Bharat Panda <bharat.b.panda@intel.com>


JeevakaPrabu (commented):
ok with the changes. Address the indentation issues with printk and commented out original code

JeevakaPrabu (approved):

PR #1417, opened 2022-04-24
add an virtual display for bootup

haowenx1234:


haowenx1234:
add an virtual display for bootup

haowenx1234:
add an virtual display for bootup

PR #1482, opened 2022-06-27
Fix for bluetooth crash

gowthambabu92:


gowthambabu92:
@bpanda24 please review.

PR #1519, opened 2022-07-18
Modify sse2/avx2 memcpy & memmove to AOSP memmove

amritaintel:
Optimization was targeted seperately for memmove
and memcpy. But this means no support for
overlapping copy when using memcpy. So modify
sse2/avx2 version of memcpy and memmove to
point to AOSP's memmove


shyjumon-n (approved):
LGTM

PR #1581, opened 2022-09-01
Optimize cbrt math api by adding vfmadd

amritaintel:
Blanket enable avx for all math APIs from
arm-optimized-routines.

This patch also reverts svml implementation
for some of the math APIs as it is found to
cause regression in Geekbench.

Tracked-On: OAM-103681


yyao3 (approved):
Looks good to me

PR #1586, opened 2022-09-06
Enable dynamic cache config

amritaintel:
This patch reads cache info from CPU.
Cache info is used in memory and string functions.

Signed-off-by: Vinay Prasad Kompella <vinay.kompella@intel.com>


amritaintel:
Dynamic cache PR has been moved to R2 ring. So closing this PR

PR #1593, opened 2022-09-07
Moving Thermal Diff patch from BSP to AOSP diff

vilasrk:
To Keep the source directory cleaner and
more readable above mentioned movement
of diff patch is done.

Tracked-On: OAM-103812
Signed-off-by: Vilas R K <vilas.r.k@intel.com>


swaroopbalan (approved):

shyjumon-n (approved):
LGTM

PR #1594, opened 2022-09-07
Revert "Add SOF firmware and topology files"

padmashree9110:
This reverts commit 6c1ba62377e47e48e4999b12f955821f3e054dd2.

Tracked-On: OAM-103646
Signed-off-by: pmandri padmashree.mandri@intel.com


balajim001 (approved):
LGTM

gkdeepa (approved):

kaushlen (approved):

JeevakaPrabu (changes requested):
Reason why the sof files are kept as bsp diff is vendor-linux-firmware content comes from open source firmware git. if we patch vendor-linux-firmware it will be lost when we take latest open source firmware

PR #1606, opened 2022-09-12
Fix phone process crash when there is no modem

JeevakaPrabu:
Phone process keeps crashing with ArrayIndexOutOfBoundsException exception.

When there is no modem, tm.getSupportedModemCount returns 0. mTelephonyCallbacks object array is created with 0 but index 1 is accessed resulting in ArrayIndexOutOfBoundsException exception.

Fix the issue by allocating and accessing mTelephonyCallbacks only when tm.getSupportedModemCount is greater than 0.

Tracked-On: OAM-103825
Signed-off-by: Jeevaka Prabu Badrappan <jeevaka.badrappan@intel.com>


shyjumon-n (approved):
LGTM

PR #1616, opened 2022-09-22
Fix drawing delay issue in cases of resize and rotation for SRIOV

lyintel:
gd_egl_draw does not flush anything when draw_submitted=false, this inludes the cases of resize and rotation, so add one time flush to fix it.

Tracked-On: OAM-104047
Signed-off-by: Lu Yang A <yang.a.lu@intel.com>


lyintel:
We have tried another fix to revert  "ui/gtk: skip any extra draw of same guest scanout blob res", commit 55f4b767f64ee0fec397c7ceebfea1473b725963, but it will cause QEMU crash after rotation. So, we choose to use current commit to fix the issue.

swaroopbalan (approved):

JeevakaPrabu (changes requested):
fix the typo "inludes -> includes "

JeevakaPrabu (approved):

PR #1665, opened 2022-11-15
Optimizing avc SW-Decoder with handwritten AVX2 intrinsic

bosepriy:
Each of the inter prediction, inverse transform and deblocking filter kernels of the H.264 decoding works by handling the 16x16 macro block of 256-bit vectors  in two operations 128-bit each in the current  SIMD support. By inferencing with hand-written code logic using AVX2 intrinsic we are able to process the same number of data with a single instruction and hence help in improving the performance of the software decoder.
The overhead(%) for cpu-cycles on the module : libcodec2_soft_avcdec has been reduced by ~15% impact when we collected the simpleperf data by playing 1024 resolution video playback on the Galley App with HW decoder disabled on Android T. Other Benchmark KPIs had no impact as this patch is related to s/w video decoding.

Tracked-On: OAM-104805
Signed-off-by: Priyanka Bose <priyanka.bose@intel.com>


shyjumon-n (approved):
LGTM

PR #1672, opened 2022-11-21
Configure scudo to use exclusive TSD

amritaintel:
Exclusive TSD reduces thread contention for allocation and improves multicore performance on Geekbench.

Signed-off-by: Vinay Prasad Kompella <vinay.kompella@intel.com>


amritaintel:
Delete this PR

PR #1673, opened 2022-11-22
Always build boot-debug.img

GangSecurity:
The image is needed in VTS test

Tracked-On: OAM-104925
Signed-off-by: Chen, Gang G <gang.g.chen@intel.com>


tanujtekriwal (commented):
I hope this patch you are not planning to merge.

PR #1677, opened 2022-11-23
Device going to offline during cts tests.

ankithbti52509:
While it remove the user, then all services which had requested for ethernet network, request for release network. After that since there is no request pending, it stop ethernet.

Here we need to take network reference count in consideration if ethernet start due to interface link state change.

Tracked-On: OAM-102974
Signed-off-by: Ankit Agrawal <ankit.agarwal@intel.com>


tanujtekriwal (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #1681, opened 2022-11-25
Revert libm patches regressing on ADL

amritaintel:
Tracked-On: OAM-105001


shyjumon-n (approved):
should bring in after fixing the ADL specific bugs.
LGTM

PR #1719, opened 2023-01-03
T OS libm avx2 optimization

soniravi31:
Signed-off-by: Soni, Ravi Kumar <ravi.kumar.soni@intel.com>


aosp_diff/preliminary/bionic/0006-bionic-libm-avx2-optimization.patch
```diff
@@ -0,0 +1,1045 @@
...
+index 1f6dca5bb..db5a6d81a 100644
+--- a/libm/upstream-freebsd/lib/msun/src/e_acos.c
++++ b/libm/upstream-freebsd/lib/msun/src/e_acos.c
+@@ -105,7 +105,8 @@ __ieee754_acos(double x)
+ 	    return 2.0*(df+w);
+ 	}
+ }
+-
++/*
+ #if LDBL_MANT_DIG == 53
```
vinaykompella, line 220: This #if is missing wen we moved the weak references to dynamic dispatch

aosp_diff/preliminary/bionic/0006-bionic-libm-avx2-optimization.patch
```diff
@@ -0,0 +1,1045 @@
...
+-END(sinh)
++END(sinh_generic)
+ # -- End  sinh
+ 
+ # Start file scope ASM
+diff --git a/libm/x86/kabylake/acos_avx2.c b/libm/x86/kabylake/acos_avx2.c
+new file mode 100644
+index 000000000..280eeaf7d
+--- /dev/null
++++ b/libm/x86/kabylake/acos_avx2.c
```
vinaykompella, line 442: Do we really need new files? Just check if this is the way we do in bionic. We have other options of using #defines and have a single .c file

aosp_diff/preliminary/external/arm-optimized-routines/0001-arm-optimized-routines-libm-avx2-optimization.patch
```diff
@@ -0,0 +1,73 @@
...
+     target: {
+diff --git a/math/log.c b/math/log.c
+index d3b7bc6..c964954 100644
+--- a/math/log.c
++++ b/math/log.c
+@@ -154,8 +154,10 @@ log (double x)
+   return eval_as_double (y);
+ }
+ #if USE_GLIBC_ABI
++#ifndef LOG_FROM_LIBM
```
vinaykompella, line 51: Use better meaningful defines

vinaykompella (changes requested):

aosp_diff/preliminary/bionic/0006-bionic-libm-avx2-optimization.patch
soniravi31: added this #if in dynamic dispatch

PR #1757, opened 2023-02-09
Patch to add av1 support entries for server streamer and encoder

weixion2:
Signed-off-by: Weixiong Zhang <weixiong.zhang@intel.com>


bsp_diff/common/vendor/intel/external/project-celadon/cloud-streaming/0001-Add-entry-point-script-for-aic-cloud-env.patch
```diff
@@ -167,10 +170,15 @@ index 0000000..19edf25
 +
 +  local renderfps_enc_enable=${ENV_RENDERFPS_ENC_ENABLE:-"false"}
 +  local hevc_enable=${HEVC_ENABLE:-"false"}
++  local av1_enable=${AV1_ENABLE:-"false"}
```
dvrogozh, line 173: That's bad options style. Problem with this style of options is that they conflict with each other. User has a chance to type in: `-hevc -av1` and one of them will take preference.

bsp_diff/common/vendor/intel/external/project-celadon/cloud-streaming/0001-Add-entry-point-script-for-aic-cloud-env.patch
```diff
@@ -167,10 +170,15 @@ index 0000000..19edf25
...
 +  local codec="264"
 +  if [[ "$hevc_enable" == "true" ]]; then
 +      codec="265"
 +  fi
++  if [[ "$av1_enable" == "true" ]]; then
```
dvrogozh, line 178: Moreover you already introduce discrepancy on your own scripts working with `-hevc -av1` typos. Here preference will go to AV1. But in https://github.com/intel-innersource/os.android.bsp.hardware-intel-android-container/pull/790 preference will go to HEVC. This happens because there you use if-else approach and here you use if-if.

dvrogozh (changes requested):

PR #1801, opened 2023-03-24
Fix for Home Button When Google Assistant is disabled.

ankithbti52509:
When user disables long press of home button option in settings to launch Google assistant, still it launch Google assistant on doing long press of Home button.

Added flag to check, if Google assistant is disabled or not.

Tracked-On: OAM-106871


shyjumon-n (approved):
LGTM

PR #1859, opened 2023-04-27
remove vhal diff patches

shivasku82:


balajim001 (changes requested):
Can you help update commit message why we are removing diff patches?

PR #1880, opened 2023-05-11
Force stop bootanimation when CarLauncher UI ready

zhou-jingyu:
The default bootanimation has all its segments' type set to 'c',wihch means "this part will play to completion, no matter what".
On ADL IVI platform, it takes less than 2 seconds from bootanimation start to boot finished (and system property 'service.bootanim.exit' set). But the default bootanimation is about 6 seconds long, which means the bootanimation will continue playing for 3~4 seconds after the system boot finished and UI ready.
This change is a workaround to force stop bootanimation when CarLauncer UI ready, so as to benchmark the time from cold boot to launcher UI correctly.

Tracked-On: OAM-109795

### DIFF
--- a/aosp_diff/celadon_ivi/packages/services/Car/0006-WA-Force-stop-bootanimation-when-CarLauncher-UI-read.patch
+++ b/aosp_diff/celadon_ivi/packages/services/Car/0006-WA-Force-stop-bootanimation-when-CarLauncher-UI-read.patch
@@ -1,0 +1,68 @@
+From 38b6c40fc80542eabeda35e185d9e84a9eb21fb8 Mon Sep 17 00:00:00 2001
+From: Zhou Jingyu <jingyu.zhou@intel.com>
+Date: Thu, 11 May 2023 12:40:25 +0800
+Subject: [PATCH] [WA] Force stop bootanimation when CarLauncher UI ready
+
+The default bootanimation has all its segments' type set
+to 'c',wihch means "this part will play to completion,
+no matter what".
+On ADL IVI platform, it takes less than 2 seconds from
+bootanimation start to boot finished (and system property
+'service.bootanim.exit' set). But the default bootanimation
+is about 6 seconds long, which means the bootanimation
+will continue playing for 3~4 seconds after the system
+boot finished and UI ready.
+This change is a workaround to force stop bootanimation
+when CarLauncer UI ready to improve the cold boot time.
+
+Tracked-On: OAM-109795
+Signed-off-by: Zhou Jingyu <jingyu.zhou@intel.com>
+---
+ .../car/SystemActivityMonitoringService.java    | 17 +++++++++++++++++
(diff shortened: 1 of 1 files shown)


aosp_diff/celadon_ivi/packages/services/Car/0006-Force-stop-bootanimation-when-CarLauncher-UI-ready.patch
```diff
@@ -0,0 +1,69 @@
...
++    private boolean mCheckLauncherStartedDone = false;
++
+     public SystemActivityMonitoringService(Context context) {
+         mContext = context;
+         mProcessObserver = new ProcessObserver();
+@@ -322,6 +325,20 @@ public class SystemActivityMonitoringService implements CarServiceBase {
+                         Slog.i(CarLog.TAG_AM, "Updating top task to: " + newTopTaskInfo);
+                     }
+                 }
++                // workaround to force stop bootanimation when CarLauncher UI ready, so
```
shyjumon-n, line 50: shall remove the ", so" from comment

aosp_diff/celadon_ivi/packages/services/Car/0006-Force-stop-bootanimation-when-CarLauncher-UI-ready.patch
```diff
@@ -0,0 +1,69 @@
...
++                // as to benchmark the cold boot time to launcher UI correctly.
```
shyjumon-n, line 51: shall be replaced with :
//to improve the cold boot time

aosp_diff/celadon_ivi/packages/services/Car/0006-Force-stop-bootanimation-when-CarLauncher-UI-ready.patch
```diff
@@ -0,0 +1,69 @@
+From bc84d40529e434ea6645dc74e6f2c26f75dcc4f4 Mon Sep 17 00:00:00 2001
+From: Zhou Jingyu <jingyu.zhou@intel.com>
+Date: Thu, 11 May 2023 12:40:25 +0800
+Subject: [PATCH] Force stop bootanimation when CarLauncher UI ready
```
shyjumon-n, line 4: since it is a workaround, we should add [WA] in patch subject.

shyjumon-n (changes requested):
please check the comments

aosp_diff/celadon_ivi/packages/services/Car/0006-Force-stop-bootanimation-when-CarLauncher-UI-ready.patch
zhou-jingyu: Just fixed, please review again

JeevakaPrabu (approved):

shyjumon-n (approved):
LGTM

PR #1883, opened 2023-05-15
Update init.rc for llkd-1.

Kunkka1988:
Add ro.llk.blacklist.process and ro.llk.blacklist.parent to avoid adbd to be killed by llkd-1 (Livelock) daemon in stress test.

Tracked-On: OAM-108924


Kunkka1988:
@YuanjunHuang, would you please help to review this patch?

Kunkka1988:
It doesn't meet the expected result. I'll update this patch later.

PR #1908, opened 2023-06-01
AVX2 implementation for memmove

vinaykompella:
Improves AnTuTu v9 "RAM Bandwidth" by 5%

Tracked-On: OAM-103726

### DIFF
--- a/aosp_diff/preliminary/bionic/0006-avx2-implementation-for-memmove-api.patch
+++ b/aosp_diff/preliminary/bionic/0006-avx2-implementation-for-memmove-api.patch
@@ -1,0 +1,645 @@
+From 945ef6a63fa5c616798333a8fbbdd029c3aabddf Mon Sep 17 00:00:00 2001
+From: ahs <amrita.h.s@intel.com>
+Date: Wed, 8 Mar 2023 04:59:50 +0000
+Subject: [PATCH] avx2 implementation for memmove api
+
+This patch includes handwritten avx2 assembly
+implementation for memmove 64-bit.
+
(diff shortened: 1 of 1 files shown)


aosp_diff/preliminary/bionic/0006-avx2-implementation-for-memmove-api.patch
```diff
@@ -0,0 +1,645 @@
...
+
+diff --git a/libc/Android.bp b/libc/Android.bp
+index e4f40b323..5d7accf88 100644
+--- a/libc/Android.bp
++++ b/libc/Android.bp
+@@ -1019,6 +1019,7 @@ cc_library_static {
+                 "arch-x86_64/kabylake/string/avx2-memchr-kbl.S",
+                 "arch-x86_64/kabylake/string/avx2-memrchr-kbl.S",
+                 "arch-x86_64/kabylake/string/avx2-memset-kbl.S",
++                "arch-x86_64/kabylake/string/avx2-memmove-kbl.S",
```
shyjumon-n, line 27: Are we using the kabylake folder name ?

shyjumon-n (commented):
Check comment and update

aosp_diff/preliminary/bionic/0006-avx2-implementation-for-memmove-api.patch
vinaykompella: Traditionally it was used, but right now there is no such need for this. However, some configuration like cache.h still exists specific to architecture, hence a separate folder with that name. With DCC that also should go away.

shyjumon-n (approved):
LGTM

PR #1914, opened 2023-06-08
Upgrade s master to Android 12L

tanujtekriwal:
Tracked-On: NA


bsp_diff/celadon_ivi/device/intel/mixins/0002-Disable-evs-hal-and-service-as-it-is-crashing-on-12L.patch
```diff
@@ -0,0 +1,30 @@
+From 92b55721da247ab7e5b2a34d3e31fcb6c8148c36 Mon Sep 17 00:00:00 2001
+From: Tanuj Tekriwal <tanuj.tekriwal@intel.com>
+Date: Wed, 31 May 2023 13:23:41 +0530
+Subject: [PATCH] Disable evs hal and service as it is crashing on 12L
```
shyjumon-n, line 4: May be we should have [WA] tag for this patch.

shyjumon-n (commented):
Don't we need mixin up patch ? 

Few comments pls check

shyjumon-n (approved):
LGTM

PR #1917, opened 2023-06-13
Update the host patches for qemu 7.2.3

yhe39:
The 0001 revert the commit 0f957c53c84d, causes qemu launch fail.
It exit(1) if audio backend failed to be found or initialized.
The 0002 revert the commit 5890258aeeba, causes qemu launch fail.
It remove the slirp submodule, since QEMU 7.1 don't support
Ubuntu 18.04 anymore.
The 0003 revert the commit 88738ea40bee, causes SRIOV boot fail.
It will cause display corrupt when changing screen color depth
since 7.0.
The 0004-0012 are the 9 patches porting from qemu 7.0.0.

Tracked-On: OAM-110647

### DIFF
--- a/host/qemu/0001-Revert-audio-exit-1-if-audio-backend-failed-to-be-fo.patch
+++ b/host/qemu/0001-Revert-audio-exit-1-if-audio-backend-failed-to-be-fo.patch
@@ -1,0 +1,100 @@
+From 2bfbfbf2f39b8e10b702e182263accb9c0b3293a Mon Sep 17 00:00:00 2001
+From: HeYue <yue.he@intel.com>
+Date: Mon, 12 Jun 2023 16:55:33 +0800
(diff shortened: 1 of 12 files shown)


feijiang1 (commented):
"It remove the remove" need duplicated "the remove" in description.
And please mention other patches are porting from qemu 7.0 patches.

feijiang1 (approved):

JeevakaPrabu (changes requested):
- Please add why we are reverting this 3 patches in detailed commit message 
- generate the patches with git format-patch --no-numbered -12 to avoid 1/12..10/12

JeevakaPrabu (approved):

YadongQi (approved):

swaroopbalan (approved):

PR #1923, opened 2023-06-15
Save app setting of "Peak Performance" to database immediately.

XuBing0:
From the code, save to database only when system shutdown, data will not be saved if the system reboot or shutdown unexpectly, so we save the data when it is changed.

Tracked-On: OAM-108706

### DIFF
--- a/aosp_diff/celadon_ivi/packages/services/Car/0008-Save-app-setting-of-Peak-Performance-to-database-imm.patch
+++ b/aosp_diff/celadon_ivi/packages/services/Car/0008-Save-app-setting-of-Peak-Performance-to-database-imm.patch
@@ -1,0 +1,32 @@
+From f57de26893837601d98215d701a77217bba3ddde Mon Sep 17 00:00:00 2001
+From: Bing Xu <bing.xu@intel.com>
+Date: Thu, 15 Jun 2023 10:51:36 +0800
+Subject: [PATCH] Save app setting of "Peak Performance" to database
+ immediately.
+
+As per the code, save to database is done only on system shutdown
+and not on system reboot or unexpected shutdown.
+
+Fix the issue by saving the data immediately upon change.
+
+Tracked-On: OAM-108706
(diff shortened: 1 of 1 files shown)


dyang23 (approved):
LGTM

aosp_diff/celadon_ivi/packages/services/Car/0008-Save-app-setting-of-Peak-Performance-to-database-imm.patch
```diff
@@ -0,0 +1,31 @@
...
+
+diff --git a/service/src/com/android/car/watchdog/CarWatchdogService.java b/service/src/com/android/car/watchdog/CarWatchdogService.java
+index dc074a636..5a5179cc2 100644
+--- a/service/src/com/android/car/watchdog/CarWatchdogService.java
++++ b/service/src/com/android/car/watchdog/CarWatchdogService.java
+@@ -436,6 +436,7 @@ public final class CarWatchdogService extends ICarWatchdogService.Stub implement
+             boolean isKillable) {
+         ICarImpl.assertPermission(mContext, Car.PERMISSION_CONTROL_CAR_WATCHDOG_CONFIG);
+         mWatchdogPerfHandler.setKillablePackageAsUser(packageName, userHandle, isKillable);
++	mWatchdogPerfHandler.writeToDatabase();
```
JeevakaPrabu, line 25: - Please use spaces instead of tabs.

JeevakaPrabu (changes requested):
please use spaces instead of tabs.

JeevakaPrabu (changes requested):
commit message recommendation with typo corrected. 

Save app setting of "Peak Performance" to database immediately.

As per the code, save to database is done only on system shutdown
and not on system reboot or unexpected shutdown.

Fix the issue by saving the data immediately upon change.

Tracked-On: OAM-108706
Signed-off-by: Bing Xu <bing.xu@intel.com>

JeevakaPrabu (approved):

PR #1934, opened 2023-06-22
Reverting the patch to fix black screen issue.

tanujtekriwal:
Reverted patch needs to be modified according to Android 12L and then submitted back reverting the old and this patch.

Tracked-On: OAM-110758


shyjumon-n (approved):
LGTM

PR #1935, opened 2023-06-23
Fix for Calendar App to Launch in Passenger zone.

ankithbti52509:
Provider packages were not installed for profile type user, So user was not able to launch following apps in passenger zone-:
Calendar, Contacts & Phone Apps.

Added provider packages for above apps in pre-installed list.

Tracked-On: OAM-106801


shyjumon-n (approved):
LGTM

tanujtekriwal (approved):
LGTM

PR #1946, opened 2023-06-27
Fix for Local Media Player in Passenger zone.

ankithbti52509:
Currently MultiDisplay launcher does not have capability to show launcher icons for media services app, as these apps does not have any launcher category, So it does not show icon in launcher app.

Querying all MediaBrowsingService apps and adding them in launcher.

Tracked-On: OAM-106796


shyjumon-n (approved):
LGTM

PR #1950, opened 2023-06-28
Add null check for intent in KeyguardManager

tanujtekriwal:
intent should not be null before starting the activity. This null check will avoid that.

Tracked-On: NA


shyjumon-n (approved):
LGTM

PR #1969, opened 2023-07-10
Make bootanimation can be interrupted by the end of boot

zhou-jingyu:
The default bootanimation has all its segments' type set to 'c',wihch means "this part will play to completion, no matter what".
On ADL IVI platform, it takes less than 2 seconds from bootanimation start to boot finished (and system property 'service.bootanim.exit' set). But the default bootanimation is about 6 seconds long, which means the bootanimation will continue playing for 3~4 seconds after the system boot finished and UI ready.
This change set some bootanimation segments to type 'p' so that the animation can be interrupted by the end of the boot, then the user can see the launcher screen a few seconds earlier. The first 3 parts are kept as 'c' and the bootanimation has a minimum duration of about 3s.

Tracked-On: OAM-110792


shyjumon-n (changes requested):
Can we control the feature with a configuration option and attached to a target ? Depending on target the timing may change ? do we need to consider this as a tunable ?

zhou-jingyu:
> Can we control the feature with a configuration option and attached to a target ? Depending on target the timing may change ? do we need to consider this as a tunable ?

The file changed here is specific to IVI targets thus has no impact on AOW or others. The purpose of this change is to help correctly benchmark the system boot time, which is common for IVI targets. It is also self-tunable in some extents (e.g. if the system takes more than 3s to set bootanim.exit=1, then the animation will last from more than 3s up to 6s). Thus can we keep it as simplified solution for benchmark for now, and add more features in the future?

JeevakaPrabu (approved):

shyjumon-n (approved):
LGTM

PR #1994, opened 2023-07-14
bionic: memmove optimization

kaushlen:


shyjumon-n (changes requested):
Please maintain the original commit information in the commit msg, both in patch and main commit.

JeevakaPrabu (approved):

PR #2021, opened 2023-07-28
no need to exec the wait_for_keymaster

syan10:


JeevakaPrabu (changes requested):
- Can this change done in base project itself instead of maintained as diff patch?
- Detailed commit message missing the issue, root cause and fix. Please follow the commit message format

syan10:
> * Can this change done in base project itself instead of maintained as diff patch?
> * Detailed commit message missing the issue, root cause and fix. Please follow the commit message format

Jeevaka, 
this patch is for celadon-ivi project, since it has the boot time consideration, so we can remove this process to reduce the boot time. For android T later, google already have a patch in vold to remove the related codes of wait_for_keymaster.

lchen43 (approved):
LGTM

PR #2023, opened 2023-08-01
Fix "Loading" dialog not hidden timely after user SWITCHING event

zhou-jingyu:
Sometimes a "Loading" dialog is seen during boot for several seconds before launcher show up. Two problems found in Car SystemUI app may lead to this issue. This patch contains two fixes for it, and improves about 3 seconds boot time KPI.

1. Fix CalledFromWrongThreadException in function UserNameViewController.updateUser(). This exception will interrupt the user SWITCHING event delivery and cause this issue.

2. Fix the race condition when checking mShowing in UserSwitchTransitionViewController.handleHide(). The race condition may lead to skipping the actual hide action and cause this issue.

Tracked-On: OAM-111369


shyjumon-n (approved):
LGTM

PR #2031, opened 2023-08-04
Fixed for crash in Multidisplay launcher.

ankithbti52509:
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.ComponentName.flattenToString()' on a null object reference at com.android.car.multidisplay.launcher.LauncherActivity.onAppPicked(LauncherActivity.java:382)

When try pinning any media app in launcher home screen, then it try to fetch component name, but for media apps component name was not set.

Tracked-On: OAM-111351


shyjumon-n (approved):
LGTM

PR #2032, opened 2023-08-04
Fixed for crash in Media component.

ankithbti52509:
Getting following error while running monkey test-: java.lang.IllegalArgumentException: Tab position is invalid: 0 at com.android.car.ui.toolbar.TabLayout.selectTab(TabLayout.java:121) at com.android.car.ui.toolbar.ToolbarControllerImpl.selectTab(ToolbarControllerImpl.java:353) at com.android.car.media.widgets.AppBarController.setActiveItem(AppBarController.java:234)

it should show tab only when navigation bar is not present. But here it is trying to show tab, even navigation bar is present and which get crash due to tab size 0.

Tracked-On: OAM-111149


shyjumon-n (approved):
LGTM

PR #2036, opened 2023-08-07
Fixed for crash in Local media player during monkey test.

ankithbti52509:

java.lang.RuntimeException: Unable to create service com.android.car.media.localmediaplayer.LocalMediaBrowserService: java.lang.IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked

Use device protected storage to Access shared preference before reboot complete.

Tracked-On: OAM-111435


shyjumon-n (approved):
LGTM

PR #2037, opened 2023-08-07
Fixed for crash in Gallery during monkey test.

ankithbti52509:
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.ActionBar.setDisplayOptions(int)' on a null object reference at com.android.gallery3d.filtershow.FilterShowActivity.loadXML(FilterShowActivity.java:352)

if any Activity can is embedded inside another activity, then embedded activity can not get action bar. Use parent activity to get action bar.

Tracked-On: OAM-111347


shyjumon-n (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2040, opened 2023-08-08
Fixed crash in DeskClock during monkey test.

ankithbti52509:
java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter event at com.android.deskclock.LabelDialogFragment$ImeDoneListener.onEditorAction(Unknown Source:7)

event parameter is currently declared as non null. it can be null also, So declaring it as null parameter.

Tracked-On: OAM-111432


shyjumon-n (approved):
LGTM

PR #2044, opened 2023-08-10
Fix crash on EmbeddedKitchenSink.

XuBing0:
Getting following error while running monkey test-: android.view.WindowManager$BadTokenException: Unable to add window android.view.ViewRootImpl$W@fa52119 – permission denied for window type 2038
at android.view.ViewRootImpl.setView(ViewRootImpl.java:1213)

Permission isn't set before app add dialog, we check it first and set it by hand

Tracked-On: OAM-111431


ankithbti52509 (approved):

shyjumon-n (approved):
LGTM

PR #2046, opened 2023-08-14
Fixed for crash in Usb Handler during monkey test.

ankithbti52509:
java.lang.RuntimeException: Unable to resume activity {android.car.usb.handler/android.car.usb.handler.UsbHostManagementActivity} : java.lang.IllegalStateException: Cannot perform this operation because the connection pool has been closed.

Using locked mechanism in code block instead of syncronized method.

Tracked-On: OAM-111150


shyjumon-n (approved):
LGTM

PR #2047, opened 2023-08-18
Fix duplicated app icons on Car Launcher

XuBing0:
Some apps include media service, Car Launcher will draw app icon twice if it isn't defined on
packages/apps/Car/libs/car-media-common/res/values/config.xml

Tracked-On: OAM-111463


ankithbti52509 (approved):

shyjumon-n (approved):
LGTM

PR #2048, opened 2023-08-20
Removing the patch as it is causing lunch conflict

iViggyPrabhu:
This patch is causing conflict in lunch. It is due to missing blob mentioned in index line of the diff patch and hence not being able to generate fake-ancestor. To fix the same, the patch has to be recreated.


swaroopbalan (approved):

JeevakaPrabu (approved):

JeevakaPrabu (changes requested):
Commit message not in line with the file removal

JeevakaPrabu (approved):

JeevakaPrabu (approved):

PR #2064, opened 2023-08-31
Fix random crash on Ximalaya app.

XuBing0:
When zooming in/out on main screen of Ximalaya app, it will crash if the point index is invalid, so check it before using.

Tracked-On: OAM-111908


shyjumon-n (approved):
LGTM

PR #2069, opened 2023-09-05
[WA] Add a patch which remove some code to fix a reboot bug

jiaxuan-guo:
This WA aims to solve the problem that [IVI_BM]The first boot after flashing will reboot twice.
This deletes some code belongs to commit a7635718c459b95c45a09c409460570e564037dc "Pass wiped and fs_type to vold to format encrypted partition".

Tracked-On: OAM-111863


GangSecurity (approved):

GangSecurity (approved):

shyjumon-n (approved):
LGTM

PR #2089, opened 2023-09-14
Enable V4L2Loopback with fops event broadcast

neofang7:
Issues: No V4L2loopback driver and device node when system boots up. Additionally, Apps opens upstream V4L2loopback device node as an output device and reports invalid capture device errors.

RootCause: V4L2loopback device node should inform user-space level server to write streams from physical camera device to loopback device before loopback open completed. Otherwise, the loopback device node is not a capture device, but an output device, which causes device open error in user space application.

Fix Details: In v4l2loopback open and close function, add netlink to broadcast the events to user space server, block open/close ops until the server starts to write video streams.

Tracked-On: OAM-112223

### DIFF
--- a/bsp_diff/celadon_ivi/kernel/linux-intel-lts2021/0002-Introduce-v4l2loopback-implementation.patch
+++ b/bsp_diff/celadon_ivi/kernel/linux-intel-lts2021/0002-Introduce-v4l2loopback-implementation.patch
@@ -1,0 +1,8271 @@
+From 1a0c34e86bdb0b05a08668fd41720641f22a22f1 Mon Sep 17 00:00:00 2001
+From: Neo Fang <neo.fang@intel.com>
+Date: Wed, 13 Sep 2023 14:46:22 +0000
+Subject: [PATCH 1/2] Introduce v4l2loopback implementation
+
+Upstream URL: https://github.com/umlaeute/v4l2loopback.git
+Code is based on the following upstream commit on main branch:
+
+commit 2fa9d6d3ea39c24fe92211fda3f9ea02698dd9d9 (HEAD)
+Author: bakedpotato191 <vladislavp@pm.me>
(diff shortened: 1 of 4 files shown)


bsp_diff/celadon_ivi/kernel/linux-intel-lts2021/0003-Enable-v4l2loopback-as-part-of-ivi-image.patch
```diff
@@ -0,0 +1,62 @@
+From 05dc9f76fc54dc4c594c5b239a171b02dae6dd80 Mon Sep 17 00:00:00 2001
```
JeevakaPrabu, line 1: this file is not needed

bsp_diff/celadon_ivi/kernel/linux-intel-lts2021/0002-Introduce-v4l2loopback-implementation.patch
```diff
@@ -0,0 +1,8271 @@
+From 1a0c34e86bdb0b05a08668fd41720641f22a22f1 Mon Sep 17 00:00:00 2001
+From: Neo Fang <neo.fang@intel.com>
```
JeevakaPrabu, line 2: please split this into 2 patch files. 1st patch file only upstream code. 2nd patch file our changes on top of upstream patch file.

Also generate the patch file using "git format-patch --no-numbered -n" where n is the number of patches that needs to be created and copied to utils folder

JeevakaPrabu (changes requested):
please add "Tests done:" to commit message. for eg:

Tests done:
- Verified boot to UI
- Verified loopback device /dev/video50 is created
-

shivasku82 (commented):
Need to move the changes under apollo specific with new structure this changes wont be picked

PR #2099, opened 2023-09-19
Enable the memtrack to ivi kernel

yhe39:
This patch is for dumping the GFX memory usage data.

Tracked-On: OAM-112297


ShenghuaLinINTEL:
this pr is ported as https://github.com/projectceladon/vendor-intel-utils/pull/2216, so it can be closed

PR #2108, opened 2023-09-27
Fix Automotive Ethernet can not boot up success

zhaoye1:
Add some delay after tx_enable.
TODO: this may needs to be fixed for a better way.

Test:
after add this patch, Automotive ethernet can
boot up success, adb connect through this interface success

Tracked-On: OAM-112426


bpanda24 (approved):

maozhong1 (approved):
LGTM

PR #2118, opened 2023-10-04
Fix for VtsHalGraphicsComposerV2_4TargetTest Case

ankithbti52509:
while running the vts module following fail to
push error is occurring-: Attempting to push dir
'VtsHalGraphicsComposerV2_4TargetTest' to an
existing device file /data/local/tmp/.

When it try to push directory, it always shows that dir already exists even if it is not there. this
was checking dir using 'ls' command and check for
'No such file or directory' output. But 'ls' command was giving always 'Invalid argument' and in that case it was always returning showing dir exists.
During 'ls' command syscall for 'prctrl' was assiging VMA name, but VMA feature was not enabled, so it was assiging default error code EINVAL.
Enabling this feature by enabling this flag-:
CONFIG_ANON_VMA_NAME

Test:
Run 'ls' command in adb shell for some file or directory which does not exists, it should give output
"No such file or Directory" instead of "Invalid argument".

Tracked-On: OAM-112486


JeevakaPrabu (approved):
Later will be moved to device/intel/mixins when proper solution to address patch conflicts is merged

shyjumon-n (approved):
LGTM

PR #2143, opened 2023-10-12
REVERTME: Increase number of patches limit  for apply_patch

alavalas1:
Increase number of patches limit  for apply_patch

As of now apply_patch only checks for 200 patches in a folder to apply "Already applied" tag. 
Increasing the same to 250 as the number of Patches in framework-base folder is more than 200.
This will enable Lunch to not try and apply already applied patches till the number of Patches in a folder reaches 250+.

Tracked-On: OAM-112678

### DIFF
--- a/apply_patch.sh
+++ b/apply_patch.sh
@@ -33,8 +33,8 @@ function apply()
             echo "Applying ${name}"
             patch=${root}/${proj}/${name}
             change_id=`grep -w "^Change-Id:" ${patch} | awk '{print $2}'`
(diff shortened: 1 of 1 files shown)


sgnanase (approved):

tanujtekriwal (commented):
Please don't increase character limit of heading above 72.

tanujtekriwal (approved):
LGTM

shyjumon-n (approved):
workaround.
LGTM

PR #2148, opened 2023-10-16
Fix for default notification sound.

ankithbti52509:
"Default notification sound" value is set 'none' in Settings->Sound option. this was due to missing file Tethys.ogg.

Added Tethys.ogg file for audio notification.

Test:
Open Settings->Sound option in Settings app and check for "Default notification sound" value. it should not be none.

Tracked-On: OAM-112674


shyjumon-n (approved):
LGTM

PR #2158, opened 2023-10-19
Remove NetworkPreferenceApp and CarRadioApp from default installed APP

jiaxuan-guo:
Remove NetworkPreferenceAPP and CarRadioApp for now since they are not used.

Tests done: Verified device boot to UI. Verified the mentioned apps are removed.

Tracked-On: OAM-112837


shyjumon-n (approved):
LGTM

PR #2166, opened 2023-10-20
Fix duplicated app icons & navigation keys issues on MD Launcher.

ankithbti52509:
MultiDisplay launcher is having following 2 issues: 1) Some apps include media service, in those few apps do not have launcher activity. these apps are getting included in launcher by querying all MediaBrowsingService apps. Now some apps are getting included two times, one based on MediaBrowsingService & second based on launcher category.
2) Navigation keys are showing in MD Launcher app for driver user. Driver user uses Car Launcher and it already is having navigation bar.

Include apps based on only launcher category, if it have launcher activity. Also Hide navigation keys in MD launcher for driver user.

Tests:
Install some media app ex. prime video and open MD launcher. it should show prime video app only one time and also MD launcher should not show navigation keys.

Tracked-On: OAM-112750 OAM-112847


shyjumon-n (approved):
LGTM

PR #2186, opened 2023-11-03
Fixed for Unnecessary app shortcut icon in MD Launcher.

ankithbti52509:
When User try to add app shortcut from MD launcher settings in Home screen, some time it also add other apps shortcut app icons as well.

It was adding all media services app shortcut while user try to add any media service app. it should add only pinned apps.

Tests:
Go to add app shortcut option in MD launcher home screen settings option and add app shortcut for local media app. it add app shortcut only for this app.

Tracked-On: OAM-113019


shyjumon-n (approved):
LGTM

PR #2205, opened 2023-11-10
Enable dynamic cache config

soniravi31:
This patch reads cache info from CPU.
Cache info is used in memory and string functions.

Change-Id: Ia1934fd4054a00b10ab3e7c6de85e0d3fd86567e


vinaykompella (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2206, opened 2023-11-13
Fix javacrash on com.android.gallery3d as CursorIndexOutOfBoundsExcep…

XuBing0:
…tion

Media file is not exist on the path, so the cursor of the path is not null but the size is zero, so need check both.

Tracked-On: OAM-113164


tanujtekriwal (commented):
Hi Bing,
We are now following Commit message format:

Commit header

Commit Explanation

Test: <To test the fix>

Tracked-On:
Signed-Off-By:


Please align your commit message to same.

shyjumon-n (approved):
LGTM

dyang23 (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2218, opened 2023-11-17
Fix sometimes wlan1 interface for Ap create failed

zhaoye1:
When enable 5G band for AP mode, found sometimes the wlan1 for AP will init failed.
Because sometimes wifi hal init before init.rc wifi interface create, wlan1 will init failed.
Fix methed is when AP open, hal wifi_virtual_interface_create run, init wlan1 again.

Tests:
reboot 5 times, every time wifi AP can open successfully

Tracked-On: OAM-113448


dyang23 (approved):
LGTM

bpanda24 (approved):
This is a temporary workaround, we should find proper way to deal with driver init in init.rc. Approving for now.

lchen43 (approved):
LGTM

PR #2219, opened 2023-11-17
Create Lunch target from caas_vbt to caas_cvmb

swaroopbalan:
This lunch target is based on caas_vbb and caas

Test done : Boot check done

Tracked-On: OAM-113471


shyjumon-n (approved):
LGTM

PR #2222, opened 2023-11-22
Fix fragment java crash on com.android.deskclock

XuBing0:
Fragment StopwatchFragment don't attach to window, so we need check it before geting resource of the Fragment.

Tracked-On: OAM-113328


ankithbti52509 (approved):

shyjumon-n (approved):
LGTM

PR #2226, opened 2023-11-23
Fixed for JavaCrash in Gallery app.

ankithbti52509:
When try to crop or edit the image in gallery app, sometime app is crashing as it try to set new action bar which already has an action bar.

Check for action bar before creating new action bar.

Tests:
Launch Gallery app and try to edit/crop image and it should work fine.

Tracked-On: OAM-113325


shyjumon-n (approved):
LGTM

PR #2233, opened 2023-11-28
Don't let SystemUpdater app access files on Udisk

ceiba1985:
Udisk will be mounted as adoptable storage, and SystemUpdater will have no permissions to access the files on Udisk directly.

Test done: SystemUpdater won't crash when Udisk is mounted as
adoptable storage

Tracked-On: OAM-113511


shyjumon-n (approved):
LGTM

PR #2249, opened 2023-12-07
Update utils commit hash in host kernel build script

JeevakaPrabu:
Updated utils commit hash in host kernel build script for v6.1.61 rebase.

Tests done:
- Host kernel build and boot.

Tracked-On: OAM-114109


shyjumon-n (approved):
LGTM

PR #2250, opened 2023-12-08
Created new target base_aaos

swaroopbalan:
This is a common target for Common Base Line

Test done : Boot check done

Tracked-On: OAM-114053


shyjumon-n (approved):
LGTM

PR #2259, opened 2023-12-14
drm/i915: Fix i915 crash issue

beckwen:
Encountered i915 crash during Monkey test

Add filelist mutex if access device file list to avoid race conditions

Tests Done:
 -Boot with iGPU
 -i915 debugfs check
 -Reboot

Tracked-On: OAM-114178


shyjumon-n (approved):
LGTM

JeevakaPrabu (approved):

PR #2260, opened 2023-12-15
Fixed for Crashes During Monkey Test.

ankithbti52509:
During monkey test run, observing following crash
1) android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
at android.app.ContextImpl.startActivity(ContextImpl.java:1088) at android.app.ContextImpl.startActivity(ContextImpl.java:1064) at android.content.ContextWrapper.startActivity(ContextWrapper.java:411) at com.google.android.car.kitchensink.UserNoiticeDemoUiService .startDialog(UserNoiticeDemoUiService.java:122)

we cannot start an activity from a non-activity context unless you pass the intent flag FLAG_ACTIVITY_NEW_TASK. If you attempt to start an activity without passing this flag.

2) java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState at android.app.FragmentManagerImpl.checkStateLoss (FragmentManager.java:1882) at android.app.FragmentManagerImpl .enqueueAction(FragmentManager.java:1905)
at android.app.BackStackRecord.commitInternal(BackStackRecord.java:688) at android.app.BackStackRecord.commit(BackStackRecord.java:646) at com.android.contacts.activities.PeopleActivity
.updateViewConfiguration(PeopleActivity.java:739)

when you use commit() it will can throw an exception if state loss occurs but commitAllowingStateLoss() saves transaction without state loss so that will doesn't throw an exception if state loss occurs. tracking this issue under-OAM-113422

Tests:

Tracked-On: OAM-113423

### DIFF
--- a/aosp_diff/base_aaos/packages/apps/Contacts/0001-Fixed-for-Crash-in-Contact-app-during-Monkey-Test.patch
+++ b/aosp_diff/base_aaos/packages/apps/Contacts/0001-Fixed-for-Crash-in-Contact-app-during-Monkey-Test.patch
@@ -1,0 +1,45 @@
+From cdd18f1ee4fe2391e1fda1e0dca8558c687dd2a9 Mon Sep 17 00:00:00 2001
+From: Ankit Agrawal <ankit.agarwal@intel.com>
+Date: Fri, 15 Dec 2023 10:02:25 +0530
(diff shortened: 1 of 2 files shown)


shyjumon-n (approved):
LGTM

PR #2272, opened 2023-12-21
Fixed for Soft Keyboard Color & key size.

ankithbti52509:
Currently soft keyboard keys are dark and not visible properly due to dark background and also key size is small.

Changed the key color and size.

Tests:
Open any Editor, it pop up soft keyboard. it has key visible poperly.

Tracked-On: OAM-114379


shyjumon-n (approved):
LGTM

PR #2283, opened 2024-01-04
Removed Telephony module in IVI.

ankithbti52509:
Since we are not supporting telephony module in AAOS, hence removing it from image.

Tests: Device is booting fine and no crash observed.

Tracked-On: OAM-113415


shyjumon-n (approved):
LGTM

PR #2285, opened 2024-01-05
Fixed for Toast message size view.

ankithbti52509:
When tap on build number in settings to enable developer option, it shows toast message which is very small in size.

Increased size of Toast message.

Tests: Enable developer option from settings and able to see toast message clearly.

Tracked-On: OAM-114463


shyjumon-n (approved):
LGTM

PR #2292, opened 2024-01-10
Fixed for Crash in com.android.phone in IVI

ankithbti52509:
Adding back telephony package in IVI, as BT module partially is dependent on this. Also check for telephony feature before fetching cell based location.

Tests: Run stability test and has not observed any crash.

Tracked-On: OAM-114701


shyjumon-n (approved):
LGTM

PR #2293, opened 2024-01-12
Fixed for boot to recovery mode after G3 boot and Cold boot.

ankithbti52509:
Tests: delay installd binding, observe no crashes and calls eventually succeed.

Tracked-On: OAM-114512


shyjumon-n (approved):
LGTM

PR #2295, opened 2024-01-16
Add deviceIds checking back

ceiba1985:
Issue: Cannot start video conference in TencentMeeting app.

RootCause: CameraId is video device number + CameraIdOffset, it is larger than the real camera numbers. App cannot get the camera info correctly by using the CameraId.

Fix: App may get the camera info by using camera array index or the CameraId. If the CameraId passed by app is smaller than the camera array size, then will treat the CameraId as the array index. If the CameraId passed by the app is larger than or equal with the camera array size, will loop through the array to find the CameraId.

Test: TencentMeeting, Camera2 and MultiCamera all work well.

Tracked-On: OAM-110679


neofang7 (commented):
LGTM

shyjumon-n (approved):
LGTM

PR #2296, opened 2024-01-16
[Statsd] Fix binder cookie memorie leak

soniravi31:
Statsd was only deleting cookies for binder deaths when the death notifier was invoked. However, it is possible for the binder to get overwritten and for the client process to never die.

The fix for now is to not call LinkToDeath and avoid cookies entirely. We lazily remove objects from the maps when we receive a dead object exception on them.

Other options and more details are at go/statsd-binder-cookie-leak.

Test: atest statsd_test
Bug: 180471118
Bug: 202499776
Change-Id: Ie6ee3176e73a8602ca906396c0afc7df332deed1


shyjumon-n (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2301, opened 2024-01-22
Fixed for crashe during Monkey test run.

ankithbti52509:
Following crash were observed
java.lang.NullPointerException: Attempt to invoke virtual method 'void com.android.systemui.statusbar.phone.NotificationPanelView Controller.setQsScrimEnabled(boolean)' on a null object reference

Check for null condition, so during monkey test run, if view is destroyed (which is very rare in manual test), then it can safely handle.

Tests:

Tracked-On: OAM-114271


shyjumon-n (approved):
LGTM

PR #2302, opened 2024-01-22
Fix null pointer dereference coverity

gowthambabu92:
Fix null pointer dereference coverity
CID 200863: (https://github.com/projectceladon/vendor-intel-utils/pull/1 of 1): Dereference null return value (NULL_RETURNS)

Do null check before accessing the network interfaces.

Tracked-On: OAM-115263
Signed-off-by: Gowtham Anandha Babu <gowtham.anandha.babu@intel.com>


JeevakaPrabu (changes requested):
Location of the file is not correct. Patch file has to be present under aosp_diff/base_aaos/<project_folder_path>/<patch file>

bpanda24 (approved):

balajim001 (approved):
LGTM

JeevakaPrabu (approved):

PR #2320, opened 2024-01-30
Fix javacrash on com.google.android.car.kitchensink

XuBing0:
It's regression of previous patch, calling package is not allowed to draw overlays and the permission of SYSTEM_APPLICATION_OVERLAY is not permitted, meanwhile the activity to handle intent "android.settings.action.MANAGE_OVERLAY_PERMISSION" is not exist on car system, only exist on phone system, so skip the dialog.

Tracked-On: OAM-115329


shyjumon-n (approved):
LGTM

PR #2336, opened 2024-02-13
Rebase diff patches on top of 5.15.147 kernel

sgnanase:
Kernel has been rebased to 5.15.147 with optimized drm tip

Tracked-On: OAM-115616


JeevakaPrabu (changes requested):
tabs are replaced with tabs in the patch files that needed conflict resolution

bsp_diff/base_aaos/vendor/linux/firmware/0003-Update-firmware-file-for-Intel-Bluetooth-AX211.patch => This is needed

PR #2351, opened 2024-02-22
Allow uknown source apks to be installed

ceiba1985:
To fix a issue reported by Coverity previously, a patch is removed but the replacement patch is forgotten to be committed. Need to commit the missed patch.

Test Done: Unknown source apks can be installed.

Tracked-On: OAM-115697


shyjumon-n (approved):
LGTM

PR #2360, opened 2024-02-26
Allow SystemUpdater app access udisk files

ceiba1985:
SystemUpdater app is used by customers to do full ota on Intel IVI devices, it reads ota packages from udisk. To read udisk files, SystemUpdater app must run in media_rw group. This patch checks the app name and the signed certificate, third party apps have no chances to run in media_rw group. And SystemUpdater app won't be shipped with the customers' final release build, so this patch should have no security impact to IVI devices.

Test done: SystemUpdater app can do the full ota successfully.

Tracked-On: OAM-115742


JeevakaPrabu (approved):

shyjumon-n (approved):
LGTM

PR #2362, opened 2024-02-26
overlayfs: Use userxattrs on supporting kernels.

ceiba1985:
In previous kernels, overlayfs stored its xattrs with a "trusted." prefix. This requires CAP_SYS_ADMIN. As a workaround, we carried out-of-tree kernel patches to bypass the security checks on these attrs.

The 5.15 kernel however has a new mount option "userxattr". When this is set, the "trusted." prefix is replaced with "user.", which eliminates the CAP_SYS_ADMIN requirement.

On kernels >= 5.15 we can use this feature and drop some of our out-of-tree patches.

Test done: Device can boot up successfully after pushing files to /vendor/etc/vintf/manifest

Tracked-On: OAM-114073


JeevakaPrabu (approved):

shyjumon-n (approved):
LGTM

PR #2364, opened 2024-02-27
Fix JavaCrash of com.android.camera2

XuBing0:
It's corner case, if removed item is mViewItems[BUFFER_CENTER] and the item is last item, the item will be null and crash will happen, so we need add protection for this case.

Test: Run monkey test and observe, no crash happen.

Tracked-On: OAM-115328


dyang23 (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2365, opened 2024-02-27
Fix crashes in Webview Browser during MonkeyTest

SaliniVenate:
Allow DiskRead and UnbufferedIo in StrictMode builder.

Tests: Run Monkey Test, No crah observed in webview

Tracked-On: OAM-114365


shyjumon-n (approved):
LGTM

PR #2368, opened 2024-03-01
WA: Fix Monkey Test Failure in Gallery App

SaliniVenate:
Monkey Test reported below error:
java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
at com.android.gallery3d.filtershow.editors
      .EditorPanel.showImageStatePanel

Use commitAllowingStateLoss() instead of commit() so that it will not throw an exception if state loss occurs.

Tracked-On: OAM-115689


shyjumon-n (approved):
LGTM

PR #2372, opened 2024-03-09
Upgrade AOSP 12L tag to 12.1.0_r11

tanujtekriwal:
This patch will upgrade the AOSP tag to r11 from r8.

Tests Done: Build and boot

Tracked-On: OAM-116549


shyjumon-n (approved):
LGTM

PR #2380, opened 2024-03-15
Fix JavaCarsh on com.android.gallery3d when clicking the menu of

XuBing0:
resetting history

User click the menu of resetting history, App should do nothing when there is no primary image.

Test: Run monkey test and observe, no crash happen.

Tracked-On: OAM-116645


shyjumon-n (approved):
LGTM

PR #2381, opened 2024-03-19
Fixed for CarLauncher issue during suspend.

ankithbti52509:
When device goes in suspend state and resume back, map place holder activity get killed.
Restart maps place holder activity, if it is not running.

Tests: adb shell; echo mem > /sys/power/state
device goes to sleep state and now press any keyboard key to wake up, Maps placeholder is blank.

Tracked-On: OAM-114651


shyjumon-n (approved):
LGTM

PR #2387, opened 2024-03-21
Disable audio codec.

manxiaoliang:
Without considering online-video, set a property in Nuplayer to dynamically disable the audio codec. In order to quickly eliminate the audio cause when video freezes occur in the future.

Tracked-On:OAM-116765


feijiang1 (approved):
LGTM

PR #2393, opened 2024-03-22
WA - Fix Monkey Test crash in DeskClock

SaliniVenate:
Below crash was seen during MonkeyTest:
java.lang.IllegalStateException: FragmentManager
is already executing transactions

Use commitNowAllowingStateLoss instead of commitAllowingStateLoss. commitNowAllowingStateLoss executes the transaction immediately on the calling thread unlike commitAllowingStateLoss which schedules the transaction to be executed asynchronously on the main thread.

NOTE: commitNowAllowingStateLoss may potentially lead to ANR if the transaction takes a long time to execute.

Tracked-On: OAM-116600


shyjumon-n (approved):
LGTM

PR #2404, opened 2024-04-01
Fix to hwcomposer dump api

JaikrishnaNemallapudi:
dumpsys SurfaceFlinger command is crashing system due to usage of uninitialized string variable's data address.
Fixed it by allocating memory before accessing address of string variable data.

Testdone: Verified with dumpsys and dumpsys SurfaceFlinger
          No crash observed.

Tracked-On: OAM-116838


balajim001 (approved):
LGTM

balajim001 (approved):
LGTM

rajucm2007 (approved):
LKGM

PR #2405, opened 2024-04-01
Fixed for Coverity issue in Car Settings app.

ankithbti52509:
There may be null pointer exception when using volume items to create volume preference seek bar.
If volume item is null, then move to next one.

Tests: Open Settings->Sound option, Volume slider is displaying and able to slide it.

Tracked-On: OAM-115481


shyjumon-n (approved):
LGTM

PR #2423, opened 2024-04-09
Reset default value for display white balance.

XuBing0:
Reset default value for display white balance as the design doc is changed. Refer to:
https://jira.devtools.intel.com/secure/Tests.jspa#/testCase/OATV-T61955

Test: Reset to default value and check it, test OK.

Tracked-On: OAM-117067


shyjumon-n (approved):
LGTM

PR #2429, opened 2024-04-11
Handle auto-suspend for btusb_sco driver

gowthambabu92:
Add support for auto-suspend in btusb sco driver, so that when the driver is not busy (no Tx/Rx), it goes to suspend state. When any Tx/Rx, its back to active.

Tests done:
Check the runtime power status of BT usb during below scenario.
1. BT connection
2. BT music streaming
3. BT file transfer
4. BT call

Tracked-On: OAM-116918


shyjumon-n (approved):
LGTM

balajim001 (approved):
LGTM

PR #2430, opened 2024-04-11
Fix BT profile initialization by multiple tasks

gowthambabu92:
[Problem]
During stability cycle tests, below error is observed. java.lang.IllegalStateException: Task appeared more than once: #1

BT profile initiatlization done twice because of timing issue, causing FATAL exception and systemUI is crashing.

[Solution]
Make the profile initialization code synchronized, so that one task can execute at a time.

Tests done:
Flash apollo-ivi image and verify BT on/off.

Tracked-On: OAM-115576


balajim001 (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2432, opened 2024-04-12
Fix JavaCrash of android.car.usb.handler

XuBing0:
Monkey test open two tasks of android.car.usb.handler at the same time, they use one sqlite database, one task has been destoried, database is closed meanwhile other task is using the database, so crash happen, add instance count to check if database is used.

TEST: run monkey test, no crash happen.

Tracked-On: OAM-116601


XuBing0:
with other patch

PR #2434, opened 2024-04-15
Fixed for crash during Monkey test run.

ankithbti52509:
Following crash were observed
java.lang.NullPointerException: Attempt to invoke interface method 'void android.service.dreams.IDreamManager.awaken()' on a null object reference at com.android.systemui.statusbar.phone.StatusBar.lambda $awakenDreams$37(StatusBar.java:3899)

Get dream service during api runtime & also check for null condition, so during monkey test run, then it can safely handle.

Tests:

Tracked-On: OAM-117071


shyjumon-n (approved):
LGTM

PR #2435, opened 2024-04-15
LocalMediaPlayer: fix sound is still output when focus is lost

libinyang:
Sometimes, playing audio with local media player in background, switch to video play in photos, audio and video sound will be mixed.

This is because there is contest between audio playback pause and audio playback switching to next song.

Tests done: smoke test Pass

Tracked-On: OAM-115548


libinyang:
@ankithbti52509  @shyjumon-n , Could you please help review this patch. Thanks.

ankithbti52509 (approved):

balajim001 (approved):
LGTM

PR #2437, opened 2024-04-15
Fixed for crash during Monkey test run.

ankithbti52509:
Following crash was observed in com.android.deskclock during monkey as android.app.RemoteServiceException$CannotPostForegroundService NotificationException: Bad notification for startForeground.

Create the notification channel with same channel id before creating the notification. Also context will be used of service, as this is service notification.

Tests:

Tracked-On: OAM-117151


shyjumon-n (approved):
LGTM

PR #2441, opened 2024-04-19
Revert "Handle auto-suspend for btusb_sco driver"

gowthambabu92:
This reverts commit d9cc7265a106cc25798c06555dff10b9241b841b.

1st time Bluetooth works normally after flashing, after we do #adb reboot, Bluetooth goes to bad state.

Revert this patch which is causing the issue.

Tracked-On: OAM-117365


shyjumon-n (approved):
LGTM

PR #2454, opened 2024-04-28
Enable AX210 and AX201 Bluetooth for Android Automotive

zhaoye1:
The BT chip is loaded as USB interface as hci0, the UsbHostManager
    is taking control of hci0 interface and doing initialization.
    Because of which Bluetooth-HAL is not able to send HCI cmds via
    hci0 interface, causing the hci_timeout_abort() crash.
    So BT is not enabled.

    Add BT chip vendor id and product id in config deny list, so that
    UsbHostManager skips the USB interface intialization for hci0.

    Tests:
    Open bluetooth in Settings app,
    bluetooth device can be found.

    Tracked-On:  OAM-118006


dyang23 (approved):
LGTM

qshi5 (changes requested):
Must we modify AOSP for feature enabling? If yes, we need to evaluate the impact after changing GSI.

qshi5 (approved):

zhy1030 (approved):

dyang23 (approved):
LGTM

lchen43 (approved):
LGTM

PR #2457, opened 2024-04-29
Add sprite plane and rotation for virtio gpu

liuhang-bit:
Sprite planes of physical pipe are exposed to guest through virtio gpu, and all the supported format and modifier and rotation capabilities are transferred to frontend driver, so that user space graphic software could easily utilize them to avoid unnecessary gpu computing resource waste.

Tracked-On: OAM-114060


liuhang-bit:
@feijiang1  please help to review

PR #2459, opened 2024-04-30
Fixed for Coverity issue in Car Service.

ankithbti52509:
Captured concurrent data modification issue.

Synchronized data modification variable.

Tests: Build is successfull and device is booting fine.

Tracked-On: OAM-118066


shyjumon-n (approved):
LGTM

PR #2465, opened 2024-05-07
Enable AIDL for audio

padmashree9110:


aosp_diff/caas/frameworks/av/0001-Not-To-Merge-Enable-AIDL.patch
```diff
@@ -0,0 +1,58 @@
...
+  */
+ static const std::array<AudioHalVersionInfo, 5> sAudioHALVersions = {
+     // TODO: remove this comment to get AIDL
+-    // AudioHalVersionInfo(AudioHalVersionInfo::Type::AIDL, 1, 0),
++    AudioHalVersionInfo(AudioHalVersionInfo::Type::AIDL, 1, 0),
+     AudioHalVersionInfo(AudioHalVersionInfo::Type::HIDL, 7, 1),
+     AudioHalVersionInfo(AudioHalVersionInfo::Type::HIDL, 7, 0),
+     AudioHalVersionInfo(AudioHalVersionInfo::Type::HIDL, 6, 0),
+     AudioHalVersionInfo(AudioHalVersionInfo::Type::HIDL, 5, 0),
+-    AudioHalVersionInfo(AudioHalVersionInfo::Type::HIDL, 4, 0),
```
libinyang, line 25: @padmashree9110 , I just saw your patch to enable AIDL. I'm curious why remove version 4, not all version?

padmashree9110: This patch is not for merge. Once AOSP rebase is done to latest, this change will be there as per my understanding. Since array limit is 5, i removed last element. We can even increase limit to 6. Anyway if latest code after rebase also doesn't have this change, then i will make limit to 6 and then merge.

PR #2478, opened 2024-05-20
Fix com.android.systemui crash during Monkey test

pugong2019:
Following crash were observed:
java.lang.NullPointerException: Attempt to invoke interface method 'void android.service.dreams.IDreamManager.awaken()' on a null object reference at com.android.systemui.statusbar.phone.StatusBar.lambda $awakenDreams$37(StatusBar.java:3901)

Re-set mDreamManager in runtime to avoid the first null init conditon.

Tests:

Tracked-On: OAM-118790


bing8deng (approved):

ankithbti52509 (approved):
LGTM

PR #2495, opened 2024-06-10
Enable sco driver for AX211

gowthambabu92:
btusb sco snd card driver is not present in caas.

Add btusb sco snd card driver to enable BT HFP call in caas.

Tests done:
1. Flash BM
2. Check btusb_sco_snd_card.ko in vendor/lib/modules
3. Check dmesg for btusb_sco driver registered
4. In adb shell #>cat proc/asound/cards, it should list btaudiosource
5. Install zoom application
6. Connect BT headset
7. Make zoom call
8. Check uplink and downlink works

Tracked-On: OAM-120501


shyjumon-n (approved):
LGTM

PR #2497, opened 2024-06-10
Fix java crash when Bluetooth MAP profile connects

gowthambabu92:
When Bluetooth MAP profile connects, Persistent bundle is accessed even before its being populated in the constructor resulting in ArrayIndexOutOfBoundsException.

Add a proper check for bundle initialization before querying for its configuration.

Tests done:
1. Flash Bare Metal
2. BT on success
3. Connect Pixel reference phone
4. Connect MAP
5. Disconnect/Connect MAP profile 20 times
6. No crash observed

Tracked-On: OAM-119041


shyjumon-n (approved):
LGTM

PR #2498, opened 2024-06-11
Enable wifi RTT and NAN features

gowthambabu92:
RTT and NAN features are not enabled in caas.

Enabled RTT and NAN for caas.

Tests done:
1. Flash BM
2. In adb shell #>pm list features | grep wifi
3. Verify RTT and wifi aware service is running from logcat

Tracked-On: OAM-120503


bpanda24 (approved):

shyjumon-n (approved):
LGTM

PR #2500, opened 2024-06-14
Fix crash in Webview Browser during MonkeyTest

SaliniVenate:
Getting below crash when running monkey test:
 java.lang.RuntimeException: StrictMode ThreadPolicy violation
 Caused by: android.os.strictmode.UnbufferedIoViolation

Allow UnbufferedIo in StrictMode builder.

Test Dome: Boot check
Tracked-On: OAM-118786


shyjumon-n (approved):
LGTM

PR #2501, opened 2024-06-17
CTS Regression issue -libnueralnetworks

alavalas1:
As present pathes are removed from codebase
due to CTS -failure in android.jni.cts.
JniStaticTest class in CtsJniTestCases module

Patches details
1)0001-Generate-avx2-versions-of-libnueralnetworks.so.patch 2)0002-Modify-IA-perf-variant-of-library-to-have-same-name-.patch

Test done: Build and boot is fine.

Tracked-On: OAM-120542


bosepriy (approved):
LGTM

shyjumon-n (approved):
Shall revert now, as blocker CTS need to be addressed.
LGTM.

PR #2503, opened 2024-06-18
Fix for VtsHalLoudnessEnhancerTargetTest Case

ankithbti52509:
while running the vts module following fail to push error is occurring: Attempting to push dir 'VtsHalGraphicsComposerV2_4TargetTest' to an existing device file /data/local/tmp/.

When it try to push directory, it always shows that dir already exists even if it is not there. this was checking dir using 'ls' command and check for 'No such file or directory' output. But 'ls' command was giving always 'Invalid argument' and in that case it was always returning showing dir exists. During 'ls' command syscall for 'prctrl' was assigning VMA name, but VMA feature was not enabled, so it was assigning default error code EINVAL. Enabling this feature by enabling this flag-: CONFIG_ANON_VMA_NAME

Test:
Run 'ls' command in adb shell for some file or directory which does not exists, it should give output "No such file or Directory" instead of "Invalid argument".

Tracked-On:


shyjumon-n (approved):
LGTM

PR #2514, opened 2024-06-24
InputDevice: Use location for idc file

ankithbti52509:
Devices that have multiple usb touchpanels attached that are all using the same product, vendor, and version ids do not have a good way to use different idc files to allow the touchpanels to be configured differently as matching is done either by product, vendor, and version ids or device name which will be the same for all of the touchpanels.

This updates the logic to search for idc files to add searching by InputDeviceIdentifier.location if no existing matches are found. The location is an identifier such as usb-xhci-hcd.1.auto-1/input1 which contains a bus id to allow configuration by which usb port the touchpanel is attached to. This will use the canonical location which replaces anything other than A-Z, a-z, 0-9, -, and _ with the _ character similar to what is done for the canonical name to allow it to be used in a path.

Tests: Verified 2 touch display by mapping location based idc files to display.

Type: New Feature
Change-Id: I30db47889f4925e8eb6522640bd8316af20ca572

Tracked-On: OAM-120439


shyjumon-n (approved):
LGTM

PR #2523, opened 2024-07-04
Dynamically disable audio codes

manxiaoliang:
Set a property in Nuplayer to dynamically disable the audio codec. In order to quickly eliminate the audio cause when video freezes occur in the future.

Tests Done
Disable audio decoding:
-setprop persist.sys.disable.audio true
-reboot
Enable audio decoding:
-setprop persist.sys.disable.audio false
-reboot

Tracked-On:OAM-116765


feijiang1 (approved):
LGTM

zhy1030 (approved):

PR #2536, opened 2024-07-12
Remove car ui related sepolicy rules from diff

ceiba1985:
This patch will be applied to device/intel/sepolicy directly.

Test done: Device can work normally.

Tracked-On: OAM-122232


shyjumon-n (approved):
LGTM

PR #2553, opened 2024-07-30
Removing support for 32bit since Android removed it

AlamIntel:
This patch will remove support for 32bit from the apilist

Tests Done: #getprop | grep abilist
	    This should not show 32bit support

Tracked-On: OAM-118704


tanujtekriwal (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2567, opened 2024-08-09
Fixed for crash in Gallery during monkey test.

XuBing0:
Fragment Activity does not support action bar by default. Adding toolbar in layout to create custom action bar.

Crash log snippet looks like below:
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.ActionBar.setDisplayOptions(int)' on a null object reference at com.android.gallery3d.filtershow.FilterShowActivity.loadXML(FilterShowActivity.java:352)

Tracked-On: OAM-122773


shyjumon-n (approved):
LGTM

shyjumon-n (approved):
LGTM

PR #2569, opened 2024-08-13
Enable BT-HFP AX210

haochuan1982:
changes:
   - Add AX210 intel controller
   - Support only NB

Tracked-On: OAM-123206


balajim001 (approved):
LGTM

PR #2573, opened 2024-08-20
Skip print job if print feature is not defined

XuBing0:
System doesn't enabel the feature "android.software.print", so we skip the print job.

Test: run monkey test, no crash happen

Tracked-On: OAM-123126


dyang23 (approved):

shyjumon-n (approved):
LGTM

PR #2578, opened 2024-08-20
Fix metadata issue when reconnect pixel

gowthambabu92:
When reconnecting pixel some times metadata is not synced. bt-stack issuing two AVRC browse channel, one as acceptor, other as initator, which is causing collision in lcid and resulting in l2cap disconnection of avrc browse channel, resulting in metadata sync issue.

Don't initiate avrc browse channel as initiator, open only as acceptor. Now browse channel is always success and metadata is seen everytime.

Tests-done:
1. Flash BM
2. Connect Pixel
3. Play Music and do adb reboot
4. Everytime the metadata should sync on home screen
5. Trail 20 times

Tracked-On: OAM-123252


JeevakaPrabu (changes requested):
initator => initiator

PR #2583, opened 2024-08-26
Allow modification of Vehicle HAL properties.

vilasrk:
- When "persist.vendor.emulate.vehiclehal" is set to false, the code checks the access permissions. If the property does not have write access, it returns an ACCESS_DENIED status.

- When persist.vendor.emulate.vehiclehal is set to true, the code allows writing to Vehicle HAL properties.

Test
  Ex: adb shell setprop persist.vendor.emulate.vehiclehal true dumpsys car_service set-property-value <prop-Id> <area Id> value. dumpsys car_service set-property-value 0x11400400 0x11400400 4

Tracked-On: OAM-123892


balajim001 (approved):
LGTM

PR #2586, opened 2024-08-29
revert patch of "Disable VP9 HDR support by default"

lsun30:
revert patch of "Disable VP9 HDR support by default"


feijiang1 (approved):
LGTM

bing8deng (approved):

PR #2590, opened 2024-09-02
Removing few WA patches

SaliniVenate:
Removing all WA patches added during bring-up. These patches were introduced to resolve build errors when building with python2. Now we have migrated to python3

Tests Done: Boot check

Tracked-On: OAM


shyjumon-n (approved):
LGTM

PR #2593, opened 2024-09-05
Remove few WA patches

SaliniVenate:
Remove few WA patches introduced during bring up
since the intel build is now moved to python3.

Test Done: Boot check on all targets

Tracked-On: OAM-124321


shyjumon-n (approved):
LGTM

PR #2594, opened 2024-09-05
Fix for coverity issue for memtrack_module

tanujtekriwal:
The memtrack_module variable was going out of scope and function was exiting without freeing it. Free memtrack_module if return nullptr.

Tests Done: Build and boot.

Tracked-On: OAM-122303


shyjumon-n (approved):
LGTM

PR #2599, opened 2024-09-09
Fix error when playback can't be resumed on LocalMediaPlayer

XuBing0:
Do not check permission during playback every time, permissions have been checked when APP start, and system don't pregrant some permissions for LocalMediaPlayer, can't resume playback if the permission check failed.

Test: play music on LocalMediaPlayer then pause and resume.

Tracked-On: OAM-123021


dyang23 (approved):
LGTM

zhy1030 (approved):

PR #2603, opened 2024-09-09
Remove nnhal and its dependent component diff patches

JeevakaPrabu:
As nnhal is going to be deprecated, remove the nnhal and its dependent projects.

Tests done:
- Android boot in GVT-d config
- adb reboot

Tracked-On: OAM-124527


shyjumon-n (approved):
LGTM

PR #2609, opened 2024-09-12
Fix static analysis issues for thermal daemon for IVI

mbajaj02:

+-------------------------------+
|Argument cannot be negative |
|Waiting while holding a lock |
|Resource leak |
|Uninitialized scalar variable |
+-------------------------------+

Tracked-On: OAM-124640


vilasrk (approved):
LGTM

balajim001 (approved):
LGTM

PR #2614, opened 2024-09-17
Enable Displaying lunch menu for A15

tanujtekriwal:
This patch will set the TARGET_RELEASE version and enable lunch menu options list.
TARGET_RELEASE needs to be updated based on new AOSP updates.

Tracked-On: OAM-121983


shyjumon-n (approved):
LGTM

PR #2615, opened 2024-09-17
CI builder TARGET_RELEASE agnostic lunch support

tanujtekriwal:
This patch will support CI builder to perform lunch without depending on the TARGET_RELEASE version of the underlying AOSP code. It will allow CI builder to provide lunch as target-variant and output will be target-release-variant sent to AOSP build system.

Tests Done: source and lunch target-variant

Tracked-On: OAM-124766


shyjumon-n (approved):
LGTM

PR #2616, opened 2024-09-18
Post power state listener broadcast calls to handler

XuBing0:
Handle power state listener broadcasts on a separate thread to prevent more than one broadcast being active at one time.

Bug: 323603048
Bug: 323603043
Test: atest CarPowerManagementServiceUnitTest

Tracked-On: OAM-124540


dyang23 (approved):
LGTM

PR #2622, opened 2024-09-19
Fix incoming message notification not shown in IVI

AmanBh0709:
When IVI is connected to another device over bluetooth then if the connected device receives SMS, IVI should also display the SMS notification.

In the latest Google provided MessengerPrebuilt apk this issue is not seen. Validation Team has already confirmed that in latest apk issue is not seen.

Tracked-On: OAM-125119


shyjumon-n (approved):
LGTM

PR #2626, opened 2024-09-19
Fixed for Deadlock in IMMS Proxy Service

ankithbti52509:
Observed deadlock while switching user.

Updating imms service based on user in seperate thread.

Tests: Switch user few times. there is no dead lock issue.

Tracked-On: OAM-122779


shyjumon-n (approved):
LGTM

PR #2627, opened 2024-09-20
Clean up the memtrack patch

yhe39:
Tracked-On: OAM-125137


JeevakaPrabu (changes requested):
please hold on the merge. will remove not supported kernels cleanup

PR #2628, opened 2024-09-20
dkms: optimize the dg2 kernel driver loading time

wangyiyanga:
bin in this patch were dummy files,
the reason was first round made bzImage needed them or compile failed.

they were replaced with right ones when out of tree build done.

finally bzImage re-build.

Tracked-On: OAM-124024


feijiang1 (approved):
LGTM

zhy1030 (approved):

PR #2630, opened 2024-09-22
Add test kernel module for system and odm DLKM partition

GangSecurity:
The vts_dlkm_partition_test vts test case will check if
 “/<system/odm>/lib/modules” is a symlink to
 “/<system/odm>_dlkm/lib/modules. If there is no module
in the partition, symlink will not be created.
So this is a WA patch to pass the VTS. The WA patch can be removed if enable GKI on X86 platform.

Test Done:
Boot
vts_dlkm_partition_test

Tracked-On: OAM-118469


dyang23 (approved):
LGTM

PR #2646, opened 2024-09-27
WA for compatibility mode apk resource load failed

tingli25:
Apps that don't support car mode may run in compatibility window mode.
  If they integrate system compomnets, they might fail to load
    certain resources upon creation.
  So, we bypass the configuration to try and locate the appropriate resources.

Tracked-On: OAM-124544


zhy1030 (approved):

tingli25:
abandon

PR #2655, opened 2024-10-04
Fix BT crash due to BT audio HAL dependency

gowthambabu92:
BT LE Audio profiles requires Bluetooth Audio
HAL AIDL interface for successful initialization.

Add Bluetooth Audio HAL AIDL interface in product.mk for successful BT LE Audio profile initialization.

Tracked-On: OAM-126100


JeevakaPrabu (changes requested):
please raise the changes in mixins project itself. If it specific to aaos, have it under ivi flag

gowthambabu92:
https://github.com/projectceladon/device-androidia-mixins/pull/2072

PR #2657, opened 2024-10-08
Fixed for SystemUI crash in Passenger Display.

ankithbti52509:
In concurrent multi-user mode, SystemUI crash has seen for passenger display. It was crashing due to missing audio configuration for passenger zone.

If audio configuartion is not present for any of zone, then it should always use PRIMARY_AUDIO_ZONE.

Tests: Enabled concurrent multi-user feature and passenger display is working fine. it is showing system ui components.

Tracked-On: OAM-126129


shyjumon-n (approved):
LGTM

PR #2662, opened 2024-10-09
[WA]Fix error on joystick operation

XuBing0:
The event of joystick will be actived on the foucus window, it is only related with display ID, but is not related with user, so force to enable it on main display.

Tracked-On: OAM-125665


dyang23 (approved):
LGTM

dyang23 (approved):
LGTM

PR #2670, opened 2024-10-14
Remove SOF License

padmashree9110:
SOF License is being removed with this patch. In android14 there is no requirement for SOF, so removing it completely.

Tracked-On: OAM-126381


balajim001 (approved):
LGTM

PR #2684, opened 2024-10-23
fix for OTA update through script failure

AlamIntel:
OTA update was failing due to sepolicy permission denial. Added file context for vendor_boot_a|b partition.

Tests done: Run command ./update_device.py caas-ota-xxx.zip

Tracked-On: OAM-126630


ceiba1985 (commented):
Advise merge this patch to device/intel/sepolicy directly

PR #2689, opened 2024-10-28
Add comment explaining the use of include_** files.

tanujtekriwal:
This patch will add comments explaining what will happen if we add include_common and include_preliminary files in a build target folder.

Tests Done: source and lunch

Tracked-On: OAM-126841


shyjumon-n (approved):
LGTM

PR #2698, opened 2024-11-06
Fixed for crash in QuickSearchBox.

ankithbti52509:
Observed NullPointerException in QuickSearchBox app Caused by: java.lang.NullPointerException
at com.android.quicksearchbox.ui.SuggestionsAdapterBase .getSuggestions(SuggestionsAdapterBase.kt:63)
at com.android.quicksearchbox.ui.DelayingSuggestionsAdapter .getSuggestions(DelayingSuggestionsAdapter.kt:98)

Removing null assert condition as it is already handled by further null check condition.

Tests: Launch QuickSearchBox app and type something in search box. Press Home button. There is no crash.

Tracked-On: OAM-126864


shyjumon-n (approved):
LGTM

PR #2701, opened 2024-11-06
Fix BT tombstone by force disable BT radio

gowthambabu92:


dyang23 (commented):
Can you submit to iasw branch as well?

PR #2729, opened 2024-11-26
ANGLE: Enable Vulkan backend for HWUI

Aakash-Sarkar:
This patch is needed to run celadon with Angle as the primary OpenGL ES driver.


Aakash-Sarkar:
Not followed proper procedures

PR #2732, opened 2024-11-26
[ANGLE]: Enable vulkan backend for HWUI

Aakash-Sarkar:
This patch is needed for System UI to work correctly when Angle is used as the OpenGL ES driver.

Tests done:

- Android boot with angle and vulkan as backend
- System UI is working correctly

Tracked-On: OAM-126014


Aakash-Sarkar:
Closing this PR. Will be substituted with another one.

PR #2748, opened 2024-12-07
Configure virtual_ab compression

bunnitha:
Set vabc algo to lz4.
Enabled vabc compression threads.
Enabled vabc compression xor.

Tests done: Build-Flash-Verify

Tracked-on: OAM-125218


shyjumon-n (approved):
LGTM

PR #2755, opened 2024-12-13
Fix video playback issue with ANGLE

Aakash-Sarkar:
This set of patches fix the crash seen during video playback when ANGLE is used as the primary OpenGL ES driver with vulkan as its backend.

Tests done:

- Android boot with angle and vulkan as backend
- Video playback working correctly
- No visual artifacts seen in the Gallery App

Tracked-On: OAM-126014


Aakash-Sarkar:
Replaced with a new PR

PR #2762, opened 2024-12-18
Add support for angle as the OpenGL ES driver

Aakash-Sarkar:
This set of patches fix the visual artifacts seen in the System UI when Angle is used as the primary OpenGL ES driver with vulkan as the backend renderer. This also fix the crash seen while running video playback with Angle enabled.

Tests done:

 - Android boot with Angle and vulkan as backend
 - System UI is working correctly without any artifacts
 - Video playback working correctly
 - No visual artifacts seen in the Gallery App

Tracked-On: OAM-126014


feijiang1 (approved):
LGTM

PR #2765, opened 2024-12-19
Fix static analysis issues for thermal daemon for caas

mbajaj02:

+---------------------------------------+
|Argument cannot be negative           |
|Resource leak                                     |
+---------------------------------------+

Tracked-On: OAM-128504


balajim001 (approved):
LGTM

PR #2782, opened 2025-01-08
Fix for Android 15 Ip Provisioning Failure

AmanBh0709:
In Android 15, Wifi is not working due to ip provisioning failure. This is because of DHCP Accept packet being filtered out by DhcpFilter from jni level.

By checking the DHCP handling logic from Android 14, which does not include the problematic macros, we found that one of the macro had wrong jump statement. Now, DHCP accept packets are not filtered and can reach the router.

Tests Done:
1. Boot android image for both caas and base_aaos with this change
2. Go to settings and Connect to Wifi from router and from Mobile Hotspot
3. Wifi Connects Succesfully

Tracked-On: OAM-129211


zhaoye1 (approved):
LGTM

balajim001 (approved):
LGTM

PR #2786, opened 2025-01-08
Enable primary audio with AIDL

padmashree9110:
This patch adds implementation for dynamic detection of sound card and HDMI Audio.

Tracked-On: OAM-129138


gkdeepa (approved):

balajim001 (approved):
LGTM

PR #2804, opened 2025-01-20
Fix CTS stability issue

SaliniVenate:
Currently the CTS is not running to completion.
During CTS run, the device is going to suspend mode, and adb devices is going offline.


SaliniVenate:
fixed issue with another patch

PR #2805, opened 2025-01-21

mbajaj02:
and deprecate KBL support.

+------------------------------+
|COPY_INSTEAD_OF_MOVE          |
|Waiting while holding a lock  |
|Resource leak                 |
|Uninitialized scalar variable |
+------------------------------+

KBL is not supported anymore, therefore deprecating.

Tracked-On: OAM-129647


vilasrk (approved):
LGTM

PR #2827, opened 2025-02-14

mbajaj02:
and deprecate KBL support.

+------------------------------------+
+------------------------------------+

KBL is not supported anymore, therefore deprecating.

Tracked-On: OAM-129601


shivasku82 (approved):
LGTM

PR #2849, opened 2025-03-17

mbajaj02:

COPY_INSTEAD_OF_MOVE          
Waiting while holding a lock  
Resource leak                 
Uninitialized scalar variable 
Argument cannot be negative   

Tracked-On: OAM-131032


RajaniRanjan:
LGTM

shivasku82 (approved):
LGTM

PR #2865, opened 2025-04-09
Update the AOSP tag to r20 for A15

tanujtekriwal:
This patch will update the aosp tag to r20

Tests Done: Build and boot

Tracked-On: OAM-131768


shyjumon-n (approved):
LGTM

PR #2875, opened 2025-04-22
Porting patch to skip print job if print feature is not defined

RajaniRanjan:
System doesn't enable the feature "android.software.print", so we skip the print job.

Test: run monkey test, no crash happen

Tracked-On: OAM-131764


mbajaj02:
LGTM

shivasku82 (approved):
LGTM

PR #2886, opened 2025-04-30
Fix for android boot failure with trusty

ARavichandra:
Android fails to boot on MTL platforms when trusty is enabled, this is caused by a vmcall to activate VTD.

After enabling VTD from the bios settings this vmcall is not required.

Tests Done:
1. Flash the image in MTL nuc and CiV
2. Android is booting fine

Tracked-On: OAM-132238


balajim001 (approved):
LGTM

PR #2887, opened 2025-04-30
Fixing lunch conflict

AlamIntel:
removing conflicting patch as patch already present in codebase.

Test Done: build and boot success.

Tracked-On: OAM-132270


shyjumon-n (approved):
LGTM

PR #2892, opened 2025-05-09
Fix Potential Null Pointer Crash in CameraActivity

NaveenVenturi1203:
Uri Null pointer is not verified before getting type

Validated URI before fetching mimetype

Tracked-On: OAM-128685


shivasku82 (approved):
LGTM

PR #2894, opened 2025-05-20
Fix the issue of MapsPlaceholderActivity disappearing abnormally

XuBing0:
When pluging and unpluging USB cable, home screen will refresh and the focus of MapsPlaceholderActivity will be lost, the activity will be set to hidden state. it's can't be shown when the activity is resumed, so we relaunch the activity when the activity has already run.

Tracked-On: OAM-132415


dyang23 (approved):

XuBing0:
patch is conflict, close it and resubmit it

PR #2904, opened 2025-06-16
Fix the error of sound output disorder

XuBing0:
System will reuse the ringtone on multi-passenger mode when the ringtone has been used, but the sould will be closed when finishing to play, the sould will be played on remote and on another zone. so we always stop old sound and start to play new sound when adjusting the sound volumn.

Tracked-On: OAM-133048


dyang23 (approved):

XuBing0:
Close as we use the patch from xinxin

PR #2905, opened 2025-06-17
Fix the wrong routing when playing feedback ringtone

xinxin-wan:
Create the MediaPlayer instance for the Ringtone object before calling play() to avoid incorrect logic being triggered during ringtone playback. If the Ringtone is not created at this point, calling play() will check whether to use a RemotePlayer. The RemotePlayer is intended for certain system apps that want to play ringtones within the SystemUI process by invoking the setRemote() interface, which delegates playback to the system server side.

Tracked-On: OAM-133048


dyang23 (approved):
nice fix

dyang23 (approved):

bing8deng (approved):

